PR #1, opened 2022-11-02
Project refactor -> from one monolith to multiple services in monorepo

lucacavallaro:


grausof:
`apps/io-func-sign-issuer/src/infra/azure/storage/document.ts ` is an empty file!

PR #3, opened 2022-11-09
core: Add PDF content validator [SFEQS-1110,SFEQS-1137]

grausof:
This PR adds the validation of the pdf files uploaded through the [pdf-lib](https://pdf-lib.js.org/) library.

#### Motivation and Context
Currently the file upload flow does not provide any control on the type and / or format therefore it is possible to upload any type of file without any control. File format control is added with this PR.

This PR also fix SFEQS-1137 bug!

Sested on simulator

#### Types of changes
- [X] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

#### Checklist:


apps/io-func-sign-issuer/src/app/main.ts
```diff
@@ -12,4 +12,5 @@ if (configOrError instanceof Error) {
   throw configOrError;
 }
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const config = configOrError;
```
grausof, line 16: I don't know if it is needed, but as it was reported by lint I added this exception! @lucacavallaro

lucacavallaro: remove this file, I forgot it :)

apps/io-func-sign-issuer/src/upload.ts
```diff
@@ -79,6 +83,10 @@ export type MoveUploadedDocument = (
   destination: UploadMetadata["documentId"]
 ) => (source: string) => TE.TaskEither<Error, string>;
 
-export type DeleteUploadMetadata = (
+export type DeleteUploadDocument = (
```
lucacavallaro, line 86: what is the purpose of this operation?

apps/io-func-sign-issuer/src/app/use-cases/validate-pdf.ts
```diff
@@ -0,0 +1,17 @@
+import { PDFDocument } from "pdf-lib";
+
+import * as TE from "fp-ts/TaskEither";
+
+import { pipe } from "fp-ts/function";
+import { toError } from "fp-ts/lib/Either";
+
+export const getPdfMetadata = (buffer: Buffer) =>
```
lucacavallaro, line 8: can this function be moved in the validate-upload use case? 
and maybe we should wrap "pdf-lib" and make it an "infra" dependency

apps/io-func-sign-issuer/src/upload.ts
grausof: To remove the uploaded file from blob storage (`uploaded` container)

apps/io-func-sign-issuer/src/app/use-cases/validate-pdf.ts
grausof: I added it as an `infra` dependency on `io-sign` since it will also be used on the user side!

PR #4, opened 2022-11-10
Add SIGNED status to DOCUMENT [SFEQS-1147]

grausof:
This PR add SIGEND status to DOCUMENT to support the case where some documents are signed and others not.

### DIFF
--- a/apps/io-func-sign-issuer/openapi.yaml
+++ b/apps/io-func-sign-issuer/openapi.yaml
@@ -503,6 +503,7 @@ components:
             - WAIT_FOR_VALIDATION
             - READY
             - REJECTED
+            - SIGNED
         metadata:
           $ref: "#/components/schemas/DocumentMetadata"
         created_at:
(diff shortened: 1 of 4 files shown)


gunzip (approved):

apps/io-func-sign-issuer/openapi.yaml
```diff
@@ -561,12 +562,34 @@ components:
...
+      allOf:
+        - $ref: "#/components/schemas/Document"
+        - type: object
+          properties:
+            uploaded_at:
+              $ref: "#/components/schemas/Timestamp"
+            url:
+              type: string
+              format: uri
+            signed_url:
```
lucacavallaro, line 575: I think that we should remove this field (`signed_url`), after the signature we should update the content of `url` field

grausof: I believe that issuer is also interested in having a reference to the unsigned file as well. Let's try to understand better 😉

lucacavallaro: Why would an issuer need a link to the source document as well? I do not understand the motivation. 

The `url` field then carries the risk of being misleading

PR #5, opened 2022-11-10
refactor: Add REJECTED status to Signature Request [SFEQS-1147]

grausof:
This PR add `REJECTED` status to Signature Request with `rejected_reason` and `uploaded_at` fields.

### DIFF
--- a/apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
+++ b/apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
@@ -33,23 +33,42 @@ export const SignatureRequestToApiModel: E.Encoder<
     id,
     dossierId: dossier_id,
     signerId: signer_id,
-    status,
     createdAt: created_at,
     updatedAt: updated_at,
     expiresAt: expires_at,
     documents,
-  }) => ({
-    id,
-    dossier_id,
-    signer_id,
-    status: toApiModelEnum(status),
-    created_at,
-    updated_at,
-    expires_at,
-    documents: documents.map(DocumentToApiModel.encode),
-    // here we have to handle the dynamic QR Code
-    qr_code_url: ["DRAFT", "READ"].includes(status)
-      ? void 0
-      : "https://place-holder.com/qr-code",
-  }),
+    ...additionals
+  }) => {
+    const commonFields = {
+      id,
+      dossier_id,
+      signer_id,
+      status: toApiModelEnum(additionals.status),
+      created_at,
+      updated_at,
+      expires_at,
+      documents: documents.map(DocumentToApiModel.encode),
+      // here we have to handle the dynamic QR Code
+      qr_code_url: ["DRAFT", "READ"].includes(additionals.status)
+        ? void 0
+        : "https://place-holder.com/qr-code",
+    };
+    switch (additionals.status) {
+      case "SIGNED": {
+        return {
+          ...commonFields,
+          signed_at: additionals.signedAt,
+        };
+      }
+      case "REJECTED": {
+        return {
+          ...commonFields,
+          reject_reason: additionals.rejectedReason,
+        };
+      }
+      default: {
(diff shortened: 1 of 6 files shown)


apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
```diff
@@ -31,23 +33,42 @@ export const SignatureRequestToApiModel: E.Encoder<
...
+      dossier_id,
+      signer_id,
+      status: toApiModelEnum(additionals.status),
+      created_at,
+      updated_at,
+      expires_at,
+      documents: documents.map(DocumentToApiModel.encode),
+      // here we have to handle the dynamic QR Code
+      qr_code_url: ["DRAFT", "READY"].includes(additionals.status)
+        ? void 0
```
gunzip, line 53:
```suggestion
        ? undefined
```

apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
```diff
@@ -31,23 +33,42 @@ export const SignatureRequestToApiModel: E.Encoder<
...
+    const commonFields = {
+      id,
+      dossier_id,
+      signer_id,
+      status: toApiModelEnum(additionals.status),
+      created_at,
+      updated_at,
+      expires_at,
+      documents: documents.map(DocumentToApiModel.encode),
+      // here we have to handle the dynamic QR Code
```
gunzip, line 51: can you elaborate more on this (within the comment). why it's handled in this way?

apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
```diff
@@ -31,23 +33,42 @@ export const SignatureRequestToApiModel: E.Encoder<
...
+      qr_code_url: ["DRAFT", "READY"].includes(additionals.status)
```
gunzip, line 52: to prent typos I guess you can use

SignatureRequestStatusEnum.DRAFT / SignatureRequestStatusEnum.READY isn't it

and/or extract the array into some typed constant

apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
```diff
@@ -31,23 +33,42 @@ export const SignatureRequestToApiModel: E.Encoder<
...
-    created_at,
-    updated_at,
-    expires_at,
-    documents: documents.map(DocumentToApiModel.encode),
-    // here we have to handle the dynamic QR Code
-    qr_code_url: ["DRAFT", "READ"].includes(status)
-      ? void 0
-      : "https://place-holder.com/qr-code",
-  }),
+    ...additionals
```
gunzip, line 40: what's this value here? "additionals" sound too much generic.
(In case I'd use `extras` or `rest` anyway)

apps/io-func-sign-issuer/src/upload.ts
```diff
@@ -17,7 +17,7 @@ export const UploadMetadata = t.intersection([
   t.type({
     id: Id,
     documentId: Document.types[0].props.id,
-    signatureRequestId: SignatureRequest.props.id,
+    signatureRequestId: SignatureRequest.types[0].props.id,
```
gunzip, line 20: I suggest to extract the type of the union into its own constant instead

apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
grausof: @lucacavallaro

lucacavallaro: it's implemented this way because it's just a placeholder (to give a preview in the API) waiting for the domain part to be developed

when the domain code is developed, there won't be any kind of application logic in this function

PR #6, opened 2022-11-13
core: Notification endpoint [sfeqs-1114]

grausof:
This PR adds the endpoint to send a signature request notification via message on the io app. 
It also implements the client for sending the message !


apps/io-func-sign-issuer/src/infra/http/decoders/signature-request.ts
```diff
@@ -18,7 +18,9 @@ import { makeRequireIssuer } from "./issuer";
 const requireSignatureRequestIdFromPath = flow(
   path("signatureRequestId"),
```
gunzip, line 19: isn't `req_id`?

packages/io-services/src/message.ts
```diff
@@ -0,0 +1,18 @@
...
+import * as E from "fp-ts/lib/Either";
+import * as TE from "fp-ts/lib/TaskEither";
+
+import { pipe } from "fp-ts/lib/function";
+
+import { NewMessage } from "@pagopa/io-functions-services-sdk/NewMessage";
+
+import { IOApiClient } from "./client";
+
+export const makesubmitMessageForUser =
```
gunzip, line 11: where is this used? (there's another method whith the same name). do we need this one?

packages/pdv-tokenizer/src/signer.ts
```diff
@@ -1,13 +1,31 @@
...
 export const makeGetSignerByFiscalCode =
-  (): GetSignerByFiscalCode => (fiscalCode) =>
+  (_client: PdvTokenizerClient) => (): GetSignerByFiscalCode => (fiscalCode) =>
     TE.right(
       O.some({
         id: `${fiscalCode}-signer` as NonEmptyString,
       })
     );
+
+// TODO: This is a mock
```
gunzip, line 21: why do we need to mock this part? isn't the pdv client working yet?

grausof: nope!

packages/io-services/src/message.ts
grausof: right! Thanks!

apps/io-func-sign-issuer/src/infra/http/decoders/signature-request.ts
grausof: No, in [function.json ](https://github.com/pagopa/io-sign/blob/0f8093b6320a73cb132115324b166f6af89b9885/apps/io-func-sign-issuer/GetSignatureRequest/function.json#L8) is `signatureRequestId`

packages/pdv-tokenizer/src/signer.ts
grausof: @filippo-tenaglia is still working on bringing it into the refactor!

packages/io-services/src/message.ts
```diff
@@ -0,0 +1,63 @@
...
+      ),
+      TE.chain((createdMessage) =>
+        pipe(
+          createdMessage,
+          E.mapLeft(() => new Error("Unable to send the message!")),
+          E.chainW((response) =>
+            response.status === 201
+              ? E.right(response.value)
+              : E.left(
+                  new Error(`An error occurred while sending the message!`)
```
grausof, line 54: @gunzip Is it ok to use the client in this way or will we have to use `IResponseType`? If so, how?

gunzip: it's ok. for more complex use case you may want to try https://github.com/gvergnaud/ts-pattern

gunzip: just note that you don't want to retry in case of 401 / 403 so you need to return two different kind of errors up to the caller (transient error in case of 500 / 429 -> retry, permanent error 401 / 401 -> do not retry)

grausof: Yes! I will enter it as well as I will use `Conflict` if the notification has already been sent. I am waiting for global error handling

apps/io-func-sign-issuer/src/istitution-registry.ts
```diff
@@ -0,0 +1,21 @@
...
+  subsciptionId: Id,
+  description: t.string,
+});
+
+export type Istitution = t.TypeOf<typeof Istitution>;
+
+export const IstitutionRegistry = t.array(Istitution);
+export type IstitutionRegistry = t.TypeOf<typeof IstitutionRegistry>;
+
+export const istitutionRegistry: IstitutionRegistry = [
```
gunzip, line 15: This must be handled with a separated PR. I guess this will be more like 
`getIstitutionByIssuerId()` that queries the database (rather than a in-memory structure)

grausof: I deleted the registry part, leaving the construction of the message compatible with this new addition!

apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
```diff
@@ -0,0 +1,103 @@
...
+      }\`\n- expiresAt: \`${
+        signatureRequest.expiresAt ? signatureRequest.expiresAt : "never"
+      }\`\n- dossier: \`${dossier.id}\`\n- docs: \`${JSON.stringify(
+        signatureRequest.documents
+      )}\`\n `,
+    },
+  });
+
+const makeMessage =
+  (getDossier: GetDossier) => (signatureRequest: SignatureRequest) =>
```
gunzip, line 40: nit: a safer way to do this is to accept a modded `SignatureRequest` type that has status = t.literal(`READY`) and notification is `t.undefined`

gunzip (approved):

apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
grausof: OK thank you. I will take care of using it!

PR #7, opened 2022-11-14
SFEQS-988: implemented the tokenizer apis to manage both calls to ret…

filippo-tenaglia:
…rieve the token by fiscal code and viceversa

Signed-off-by: Filippo Tenaglia <filippo.tenaglia@pagopa.it>


grausof:
@filippo-tenaglia I made a small change to the code to not read the key from the env every time!

gunzip (approved):

PR #10, opened 2022-11-15
feat(io-func-sign-issuer): Add Info endpoint [SFEQS-1155]

lucacavallaro:
Info endpoint for healtcheck


apps/io-func-sign-issuer/src/infra/azure/functions/info.ts
```diff
@@ -0,0 +1,27 @@
...
+import * as azure from "@pagopa/handler-kit/lib/azure";
+
+import { success } from "@internal/io-sign/infra/http/response";
+
+import * as TE from "fp-ts/lib/TaskEither";
+import { pipe, identity } from "fp-ts/lib/function";
+
+import * as t from "io-ts";
+
+const makeInfoHandler = () => {
```
gunzip, line 11: What about trying to load the configuration and throw in case of errors?
(should we consider the function as unhealty in case in order to receive alarms?)

PR #13, opened 2022-11-15
Fix: prevent generation of Upload URL (Sas Token) when the SR is not DRAFT [SFEQS-1124]

lucacavallaro:
#### List of Changes

1. Add a "guard" in the `getUploadUrl` use case that checks the status of the `Signature Request`

#### Motivation and Context

The Issuer should not upload new files after the SR is marked as `READY`

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/src/app/use-cases/get-upload-url.ts
+++ b/apps/io-func-sign-issuer/src/app/use-cases/get-upload-url.ts
@@ -1,4 +1,5 @@
 import { Document } from "@internal/io-sign/document";
+import { ActionNotAllowedError } from "@internal/io-sign/error";
 import { pipe } from "fp-ts/lib/function";
 import * as TE from "fp-ts/lib/TaskEither";
 import { SignatureRequest } from "../../signature-request";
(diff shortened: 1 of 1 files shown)


grausof (approved):
LGTM

apps/io-func-sign-issuer/src/app/use-cases/get-upload-url.ts
```diff
@@ -19,8 +20,14 @@ export const makeGetUploadUrl =
   ({ signatureRequest, documentId }: GetUploadUrlPayload) =>
     pipe(
       signatureRequest,
-      newUploadMetadata(documentId),
-      TE.fromEither,
+      TE.fromPredicate(
+        (req) => req.status === "DRAFT",
```
gunzip, line 24: Another way to do this is to have a dedicated type (`ReadySignatureRequest`) and take it as input.
In this way this you cannot call `GetUploadUrl` with a wrong signature request. 

You probably want this at the API level (openapi.yaml) creating a base type / schema `SignatureRequestBase` without the `status` property. Then several different types for each status 
using `allOf`.

In this way you obtain validation for free (it comes from the OpenAPI expected payload).

What about it?

lucacavallaro: We tried that way (you can see the history of the OpenAPI spec in the old repository), but that "tagged union" approach didn't work well with our code-gen tool :( 

But for sure we can split the domain structure of the SR, as it is the layer involved in this specific scenario! 

I'll note it for future improvement!

lucacavallaro: We already tried the "tagged union" approach at the API level (with separate SR states etc.), but didn't work well with our code-gen tool. 

So, for now, we rolled back to the “flat structure” approach, which is not type-safe, but easier to use.

Regarding the domain level, I think we have to improve the current structure, but it is not required for the changes introduced with this PR.

PR #16, opened 2022-11-16
Add workflow to label automatically the PR based on the changed package

lucacavallaro:
#### List of Changes

1. Add a new `GitHub Workflow`, `.github/workflows/label.yml` (https://github.com/actions/labeler)


gunzip (approved):

.github/labeler.yaml
```diff
@@ -0,0 +1,4 @@
+io-func-sign-issuer: apps/io-func-sign-issuer/**/*
```
gunzip, line 1: it looks like the action is looking for labeler.yml (not labeler.yaml)

lucacavallaro: great! where you found out?

gunzip: in the action output

lucacavallaro: 🤦 🤦 🤦

PR #18, opened 2022-11-23
fix: Lint fix [SFEQS-1161]

grausof:
this PR performs a lint fix on the entire basecode!

### DIFF
--- a/packages/pdv-tokenizer/.eslintrc.json
+++ b/packages/pdv-tokenizer/.eslintrc.json
@@ -3,7 +3,7 @@
     "es2021": true,
     "node": true
   },
-  "ignorePatterns": ["Dangerfile.js", "src/models/requestTypes.ts"],
+  "ignorePatterns": ["Dangerfile.js", "src/models/"],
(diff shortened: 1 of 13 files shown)


packages/pdv-tokenizer/.eslintrc.json
```diff
@@ -3,7 +3,7 @@
     "es2021": true,
     "node": true
   },
-  "ignorePatterns": [],
+  "ignorePatterns": ["Dangerfile.js", "src/models/requestTypes.ts"],
```
gunzip, line 6:
```suggestion
  "ignorePatterns": ["Dangerfile.js", "src/models/"],
```

grausof: In this way I've a strange error with lint: `You are linting "src/models", but all of the files matching the glob pattern "src/models" are ignored.`

gunzip: you must keep the final slash

gunzip (approved):

PR #19, opened 2022-11-23
fix: [SFEQS-1161] fix lint tasks in package.json

gunzip:


lucacavallaro:
what is the purpose of this change?

gunzip:
@lucacavallaro if you don't enclose the glob pattern into quotes shell expansion take place and the task fail if you have excluded directories that are subdir of src

lucacavallaro (approved):

PR #21, opened 2022-11-24
feat(user): [SFEQS-1131, SFEQS-1111] Create filled document 

grausof:
This PR implements the createFilledDocument endpoint. 
Also adds:
- The PDF infrastructure for filling out forms as internal package
- Upload to blob storage
- Trigger from queue

### Infra
This PR requires an infrastructure change. 
- In particular, the container on CosmosDb: `filled-modules` 
- A queue on the storage: `waiting-for-documents-to-fill`

Tested in simulator

#### Types of changes
- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

#### Checklist:


packages/io-sign/src/infra/pdf.ts
```diff
@@ -1,12 +1,15 @@
-import { PDFDocument } from "pdf-lib";
+import { PDFDocument, PDFForm } from "pdf-lib";
```
lucacavallaro, line 1: Maybe we can move this module in a dedicated package? Such as "@internal/pdf-something"?

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -0,0 +1,91 @@
...
+  name: string;
+};
+
+// these types define the fields inside the PDF file to be enhanced
+// TODO: These are not yet the real parameters. Pending communication from the QTSP
+type NameField = Field & { fieldName: "name" };
+type FamilyNameField = Field & { fieldName: "surname" };
+type EmailField = Field & { fieldName: "email" };
+type FiscalCodeField = Field & { fieldName: "CF" };
+type Fields = [NameField, FamilyNameField, EmailField, FiscalCodeField];
```
lucacavallaro, line 28: Why are you using a fixed-size liet instead of a Record?

lucacavallaro (changes requested):

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -0,0 +1,91 @@
...
+      signer.id,
+      getFiscalCodeBySignerId,
+      TE.chain(
+        TE.fromOption(
+          () =>
+            new EntityNotFoundError("Fiscal code not found for this signer!")
+        )
+      ),
+      TE.chain((fiscalCode) => {
+        const httpApiFetch = agent.getHttpFetch(process.env);
```
lucacavallaro, line 52: "process.env" must be accessed only from "infrastructure" layer

This agent should be passed via fn arguments

lucacavallaro (changes requested):

gunzip (changes requested):

apps/io-func-sign-user/openapi.yaml
```diff
@@ -0,0 +1,186 @@
...
+paths:
+  /qtsp/clauses/filled_document:
+    post:
+      operationId: createFilledDocument
+      summary: Create a filled document for QTSP
+      tags:
+        - QTSP Clauses
+      parameters:
+        - in: header
+          name: signer_id
```
gunzip, line 16:
```suggestion
          name: x-iosign-signer-id
```

apps/io-func-sign-user/src/infra/azure/cosmos/errors.ts
```diff
@@ -0,0 +1,20 @@
+import { CosmosErrors } from "@pagopa/io-functions-commons/dist/src/utils/cosmosdb_model";
```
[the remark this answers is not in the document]
grausof: ok, I'll do it!

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
```diff
@@ -0,0 +1,120 @@
...
+} from "../../../app/use-cases/create-filled-document";
+import { makeRequireSigner } from "../../http/decoder/signer";
+import { CreateFilledDocumentBody } from "../../http/models/CreateFilledDocumentBody";
+import { FilledDocumentToApiModel } from "../../http/encoder/filled-document";
+import { FilledDocumentDetailView } from "../../http/models/FilledDocumentDetailView";
+
+import { getConfigFromEnvironment } from "../../../app/config";
+import { makeUploadFilledDocument } from "../storage/filled-document";
+
+/* TODO: This function will have to be asynchronous.
```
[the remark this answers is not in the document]
grausof: I converted it to DRAFT to add this implementation!

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
grausof: there must be only those 4 values in the array!

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
grausof: `agent.getHttpFetch` comes from  "@pagopa/ts-commons" and the signature is: `getHttpFetch: (env: NodeJS.ProcessEnv, extraOptions?: agentkeepalive.HttpOptions) => typeof fetch`

I will pass the whole configured  fetch as an argument!

apps/io-func-sign-user/openapi.yaml
```diff
@@ -0,0 +1,186 @@
...
+      description: A date-time field in ISO-8601 format and UTC timezone.
+      x-import: "@pagopa/ts-commons/lib/dates"
+      example: "2018-10-13T00:00:00.000Z"
+
+    Id:
+      type: string
+      description: Entity Id
+      format: NonEmptyString
+      example: 01ARZ3NDEKTSV4RRFFQ69G5FAV
+      x-import: "@pagopa/ts-commons/lib/strings"
```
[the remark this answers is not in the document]
grausof: so we will have to change also on the issuer openapi? @lucacavallaro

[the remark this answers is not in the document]
lucacavallaro: @gunzip why you suggest to remove this import? this will probably break our "api layer"

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
lucacavallaro: here you should use the same pattern adopted in the rest of the codebase: port & adapters

-> a `type` that represent this operation, declared in domain and used in `use-case` and a concrete implementation in `infra` with the configured `fetch`

apps/io-func-sign-user/openapi.yaml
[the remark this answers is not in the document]
gunzip: if you use `minLenght: 1` it's automatically imported

apps/io-func-sign-user/src/infra/azure/cosmos/errors.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const toCosmosDatabaseError = (e: CosmosErrors) => {
+  switch (e.kind) {
+    case "COSMOS_ERROR_RESPONSE":
+      return new CosmosDatabaseError(e.error.message);
+    case "COSMOS_EMPTY_RESPONSE":
+      return new CosmosDatabaseError("Empty response.");
+    case "COSMOS_DECODING_ERROR":
+      // TODO: show details about validation
```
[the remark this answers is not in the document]
grausof: This file derives from the package Issuer. I think we will have to do a review of all the TODOs later and fix the things that are wrong!

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -0,0 +1,91 @@
...
+  documentUrl: string;
+  email: EmailString;
+  familyName: string;
+  name: string;
+};
+
+// these types define the fields inside the PDF file to be enhanced
+// TODO: These are not yet the real parameters. Pending communication from the QTSP
+type NameField = Field & { fieldName: "name" };
+type FamilyNameField = Field & { fieldName: "surname" };
```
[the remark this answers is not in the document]
grausof: I used family_name as it does in the rest of IO. surname in this case is just a mock but it might change according to the module provided by the QTSP. Could also be "cognome"

apps/io-func-sign-user/src/infra/http/fetch-timeout.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import {
+  AbortableFetch,
+  setFetchTimeout,
+  toFetch,
+} from "@pagopa/ts-commons/lib/fetch";
+
+import { Millisecond } from "@pagopa/ts-commons/lib/units";
+
+export const makeFetchWithTimeout = (
```
gunzip, line 11: I suggest to use another argument for `makeFetchWithTimeout` which default to `process.env` (this facilitate testing). Generally, avoid inner access to global variables.

gunzip (changes requested):

apps/io-func-sign-user/FillDocument/function.json
```diff
@@ -0,0 +1,17 @@
...
+  "disabled": true,
+  "bindings": [
+    {
+      "name": "documentsToFill",
+      "type": "queueTrigger",
+      "queueName": "waiting-for-documents-to-fill",
+      "connection": "StorageAccountConnectionString"
+    }
+  ],
+  "scriptFile": "../dist/fill-document.js",
```
gunzip, line 11: I'd rename this to `CreateFilledDocument` which is the name used into API specs

apps/io-func-sign-user/FillDocument/function.json
```diff
@@ -0,0 +1,17 @@
...
+  "retry": {
+        "strategy": "fixedDelay",
+        "maxRetryCount": 4,
+        "delayInterval": "00:00:10"
```
gunzip, line 15:
```suggestion
        "delayInterval": "00:00:5"
```

apps/io-func-sign-user/openapi.yaml
```diff
@@ -0,0 +1,188 @@
...
+      description: A date-time field in ISO-8601 format and UTC timezone.
+      x-import: "@pagopa/ts-commons/lib/dates"
+      example: "2018-10-13T00:00:00.000Z"
+
+    Id:
+      type: string
+      description: Entity Id
+      minLength: 1
+      example: 01ARZ3NDEKTSV4RRFFQ69G5FAV
+      x-import: "@pagopa/ts-commons/lib/strings"
```
gunzip, line 145:
```suggestion
```

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -0,0 +1,74 @@
...
+  email: EmailString,
+  familyName: NonEmptyString,
+  name: NonEmptyString,
+});
+
+export type CreateFilledDocumentPayload = t.TypeOf<
+  typeof CreateFilledDocumentPayload
+>;
+
+/* This function returns only the callback url of the filled document without creating it.
```
gunzip, line 30: So why it's called 'makeCreateFilledDocument'?

gunzip, line 30:
```suggestion
/** Create and return the storage path (URL) for the ToS document.
 *
 * The caller of this API is expected to poll on it since the endpoint will return 404 until the ToS document gets processed.
```

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -0,0 +1,74 @@
...
+ * It also writes on a queue the information necessary to start the module creation function via the trigger queue
```
gunzip, line 31:
```suggestion
```

apps/io-func-sign-user/src/app/use-cases/fill-document.ts
```diff
@@ -0,0 +1,87 @@
...
+
+// these types define the fields inside the PDF file to be enhanced
+// TODO: These are not yet the real parameters. Pending communication from the QTSP [SFEQS-1164]
+type NameField = Field & { fieldName: "name" };
+type FamilyNameField = Field & { fieldName: "surname" };
+type EmailField = Field & { fieldName: "email" };
+type FiscalCodeField = Field & { fieldName: "CF" };
+type Fields = [NameField, FamilyNameField, EmailField, FiscalCodeField];
+
+// This function downloads the pdf form, compiles it and uploads it to blobStorage
```
gunzip, line 34:
```suggestion
/** Downloads the ToS pdf form, compiles and stores the filled document.
```

gunzip: Do not leak implementaion details (blobSorage) since they're subject to changes. Moreover, use `/** */` for methods documentation and `//` for inline comments.

apps/io-func-sign-user/src/app/use-cases/fill-document.ts
```diff
@@ -0,0 +1,87 @@
...
+            fieldValue: fiscalCode,
+          },
+        ];
+
+        return pipe(
+          TE.tryCatch(() => fetchWithTimeout(documentUrl), E.toError),
+          TE.chain((response) => TE.tryCatch(() => response.blob(), E.toError)),
+          TE.chain((blob) => TE.tryCatch(() => blob.arrayBuffer(), E.toError)),
+          TE.map((arrayBuffer) => Buffer.from(arrayBuffer)),
+          TE.chain((buffer) => pipe(fields, populatePdf(buffer))),
```
gunzip, line 83: If you invert arguments in `populatePdf` this may become 

```suggestion
          TE.chain(populatePdf(fields)),
```

as a general rule you should put configuration arguments (ie. `fields`) at first, then  data arguments (`buffer`) at last.

apps/io-func-sign-user/src/infra/azure/functions/fill-document.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+const config = configOrError;
+
+const pdvTokenizerClient = createPdvTokenizerClient(
+  config.pagopa.tokenizer.basePath,
+  config.pagopa.tokenizer.apiKey
+);
+
+const filledContainerClient = new ContainerClient(
```
gunzip, line 64: consider that these clients must be singletons (so maybe it's worth to extract it into another module in order to be shared by functions?)

packages/pdf-handler/src/pdf.ts
```diff
@@ -0,0 +1,78 @@
...
+          ? new EntityNotFoundError(e.message)
+          : new EntityNotFoundError(
+              "An error occurred while attempting to access the pdf fields."
+            )
+    ),
+    E.map((textField) => textField.setText(field.fieldValue))
+  );
+
+/* fill out the pdf form with the fields inside
+ * pdfFields and return the filled pdf in base64 */
```
gunzip, line 65: this is clear reading the code and can be removed
```suggestion
```

packages/pdv-tokenizer/src/signer.ts
```diff
@@ -80,7 +80,7 @@ export const makeGetFiscalCodeBySignerId =
               default:
                 return E.left(
                   new HttpBadRequestError(
-                    `An error occurred while connecting to the tokenizer to get Fiscal Code!`
+                    `An error occurred to retrieve the fiscal code from the signerId`
```
gunzip, line 83:
```suggestion
                    `The attempt to get signer's fiscal code from its identifier failed.`
```

apps/io-func-sign-user/FillDocument/function.json
grausof: `CreateFilledDocument` exists and is the function that takes care of the implementation of the openapi specification. `FillDocument` is the function that is triggered by a queue to fill the PDF fields.

apps/io-func-sign-user/src/app/use-cases/fill-document.ts
gunzip: https://kyleshevlin.com/just-enough-fp-argument-order

grausof: OK thank you. I will try in the future!

apps/io-func-sign-user/src/infra/azure/functions/fill-document.ts
grausof: Yes I agree. It was done this way as for the whole codebase. I saw that there is a dedicated PR so let's fix it apart.

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
grausof: throughout the codebase it was used as a convention to prepend `make` to the function name of the use case.

gunzip (approved):

lucacavallaro (approved):

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
lucacavallaro: right now we are using this naming convention

`new` prefix for "record" constructors -> `newSignatureRequest`, `newSigner`, `newIssuer`
`make` prefix for handler/use-case constructors/adapter constructors -> `makeGetSignatureRequest`

maybe in the future we can adopt `new` as prefix for all kind of constructors

gunzip: I ain't asking about the "make" parte but the "CreateFilledDocument" since looking at the comment it seems that this method does not create a filled document.

grausof: yes, I renamed it to: `makeCreateFilledDocumentUrl`

manuelcoppotelli:
Depends on pagopa/io-infra#305

PR #24, opened 2022-11-26
chore: [SFEQS-1172] Add GitHub action to apply conventional labels

gunzip:
#### List of Changes

Added GitHub workflow to automatically apply conventional labels on Pull Request.

#### Motivation and Context

This change let group Pull Requests by intent and generate nice GitHub release notes.

### DIFF
--- a/.github/release.yaml
+++ b/.github/release.yaml
@@ -7,7 +7,8 @@ changelog:
   categories:
     - title: Technical debt
       labels:
-        - technical-debt
+        - ci
+        - refactor
         - chore
         - test
     - title: Breaking changes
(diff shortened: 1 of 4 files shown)


grausof (approved):

.github/release.yaml
```diff
@@ -0,0 +1,28 @@
+changelog:
+  exclude:
+    labels:
+      - ignore-for-release
+    authors:
+      - dependabot
+  categories:
+    - title: Technical debt
```
lucacavallaro, line 8: I wouldn't make a category for "technical debt", in my opinion it should all be addressed in the tags that are used daily (https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)

gunzip: this is a mapping between PR labels and human readable relase notes group titles (labels as "chore" or "ci" don't fit well in this case).

PR #25, opened 2022-11-27
ci: add "code review" workflow [SFEQS-948]

lucacavallaro:


grausof:
Cana we add also [danger](https://github.com/pagopa/danger-plugin)?

.github/workflows/code-review.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+    name: Build and Test
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
```
grausof, line 20: You need to use the commit hash

```suggestion
        uses: actions/checkout@bf085276cecdb0cc76fbbe0687a5a0e786646936
```

.github/workflows/code-review.yaml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          fetch-depth: 2
+
+      - name: Cache turbo build setup
+        uses: actions/cache@v3
```
grausof, line 25: ditto

.github/workflows/code-review.yaml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-turbo-
+
+      - name: Setup Node.js environment
+        uses: actions/setup-node@v3
```
grausof, line 33: ditto

grausof (changes requested):

.github/workflows/code-review.yaml
lucacavallaro: Why?

lucacavallaro:
@grausof I didn't include Danger because now we have "swarmia" 

If we want to perform specific checks with Danger we can create a new workflow separated from this one

.github/workflows/code-review.yaml
gunzip: you can use plain version in case of trusted action (these come from github itself)

.github/workflows/code-review.yaml
grausof: When using third-party actions, pin the version with a commit hash rather than a tag to shield your workflow from potential supply-chain compromise.

grausof (approved):

.github/workflows/code-review.yaml
lucacavallaro: Ok, since in this workflow I didn't use third-party actions I'll leave the plain version numbers

gunzip:
just ensure we have code coverage artifacts in the end (we need to setup sonarqube integration)

lucacavallaro:
> just ensure we have code coverage artifacts in the end (we need to setup sonarqube integration)

Ok! Tomorrow I'll see how we can integrate Sonarqube. Right now this is just a porting of the previous code review pipeline.

gunzip (approved):

PR #26, opened 2022-11-27
fix: [SFEQS-1173] Make Azure clients singletons

gunzip:
#### List of Changes

Make Azure clients singletons.

#### Motivation and Context

Any PaaS client (cosmos, storage, api) must be a singleton in the context of Azure functions otherwise one connection per function is created and since outbound connections have an hard limit it's better to keep them at minimum.

It has not.


gunzip:
@lucacavallaro this is probably covered by you're refactor unsing a global "main", but just in case...

lucacavallaro:
Yes, I'll close this PR because the new `main` entry point take care of the setup of this backing services (without using "global modules")

PR #29, opened 2022-11-29
chore(user): [SFEQS-1183] Change openapi

grausof:
This PR modifies the user-side openapi specifications to make them compatible with the `io-backend` generator

### DIFF
--- a/apps/io-func-sign-user/openapi.yaml
+++ b/apps/io-func-sign-user/openapi.yaml
@@ -15,7 +15,10 @@ paths:
         - in: header
           name: x-iosign-signer-id
           schema:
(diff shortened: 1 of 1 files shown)


lucacavallaro (approved):

gunzip:
remember that we should verify the format of these ID to avoid path traversal (../../). You can use a PatternString type.

grausof:
> remember that we should verify the format of these ID to avoid path traversal (../../). You can use a PatternString type.

Ok, we have to do it also on the Issuer side!

gunzip:
yes, a Jira issue it's enough by now.

PR #35, opened 2022-12-01
feat(user): [SFEQS-1128] Add QTSP client and implement getQtspClauses

grausof:
This PR adds the client for the QTSP and implements the first getQtspClauses endpoint!

Tested in local 

#### Checklist:


apps/io-func-sign-user/src/infra/qtsp/client.ts
```diff
@@ -0,0 +1,107 @@
...
+            },
+          }),
+        E.toError
+      ),
+      TE.chainW((response) =>
+        is2xx(response)
+          ? TE.tryCatch(() => response.json(), E.toError)
+          : TE.left(
+              new HttpBadRequestError(
+                `An error occurred to retrieve the auth token from QTSP | ${response.status}`
```
gunzip, line 99:
```suggestion
                `The attempt to retrieve the the auth token from the QTSP failed: ${response.status}`
```

apps/io-func-sign-user/src/infra/qtsp/config.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import * as RE from "fp-ts/lib/ReaderEither";
+import { sequenceS } from "fp-ts/lib/Apply";
+import { readFromEnvironment } from "@internal/io-sign/infra/env";
+
+export const QtspConfig = t.type({
+  basePath: t.string,
+  username: t.string,
+  password: t.string,
+  requestTimeout: t.number,
```
gunzip, line 11:
```suggestion
  requestTimeoutMs: t.number,
```

gunzip: always put unit of measure in variable names

apps/io-func-sign-user/src/infra/qtsp/config.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const getQtspConfigFromEnvironment: RE.ReaderEither<
+  NodeJS.ProcessEnv,
+  Error,
+  QtspConfig
+> = sequenceS(RE.Apply)({
+  basePath: readFromEnvironment("QtspApiBasePath"),
+  username: readFromEnvironment("QtspUsername"),
+  password: readFromEnvironment("QtspPassword"),
+  requestTimeout: RE.right(10000),
```
gunzip, line 24:
```suggestion
  requestTimeout: RE.right(5000),
```

gunzip:
see https://github.com/pagopa/io-sign/pull/36

apps/io-func-sign-user/src/infra/qtsp/config.ts
lucacavallaro: maybe in the future we can encode this information at type level: `requestTimeout: Millisecond` instead of `requestTimeout: number`

gunzip (approved):
lgtm

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -0,0 +1,98 @@
...
+              },
+            }),
+          E.toError
+        )
+      ),
+      TE.filterOrElse(
+        isSuccessful,
+        () => new Error("The attempt to get Namirial clauses failed.")
+      ),
+      TE.chain((response) => TE.tryCatch(() => response.json(), E.toError)),
```
lucacavallaro, line 80: we should keep our code DRY

not urgent, but there is a way to avoid the repetition of `filterOrElse(isSuccessful) -> response.json()` for each call?

lucacavallaro (changes requested):

apps/io-func-sign-user/src/infra/namirial/client.ts
grausof: https://github.com/pagopa/io-sign/pull/36#discussion_r1037930912

lucacavallaro (approved):

PR #36, opened 2022-12-01
refactor: [SFEQS-1128] refactor Namirial client to avoid stateful class

gunzip:
#### List of Changes

Refactor of Namirial client.

#### Motivation and Context

1. Simplify code
1. Avoid stateful class
1. Avoid promise rejection(s)
1. Avoid access to infra layer from use-case


gunzip:
@grausof I've updated this branch, take a look

grausof:
> @grausof I've updated this branch, take a look

I agree with this implementation but @lucacavallaro suggested not to use `fp-ts` in the client but to return the promises.

gunzip:
I suppose he said that to avoid mixing stateful classes with function code. this new impl does not use classes at all

apps/io-func-sign-user/src/infra/azure/functions/get-qtsp-clauses-metadata.ts
```diff
@@ -10,28 +10,30 @@ import { error, success } from "@internal/io-sign/infra/http/response";
...
+export const makeGetQtspClausesMetadataFunction = (config: NamirialConfig) =>
+  pipe(
+    createHandler(
+      decodeHttpRequest,
+      () =>
+        pipe(
+          getQtspClausesWithToken(config),
+          TE.map(NamirialClausesToQtspClauses.encode)
+        ),
+      error,
```
grausof, line 35:
```suggestion
    createHandler(
      decodeHttpRequest,
      () =>
        pipe(
          getQtspClausesWithToken(config),
          TE.map(NamirialClausesToQtspClauses.encode)
        ),
      error,
```

```suggestion
    createHandler(
      decodeHttpRequest,
      () =>
        pipe(
          getQtspClausesWithToken(config),
          TE.mapLeft((e) => new HttpBadRequestError(e.message)),
          TE.map(NamirialClausesToQtspClauses.encode)
        ),
      error,
```
In this way, in the event of an error, we will have a left of type Error which is converted into a 500 internal server error, losing the message. I suggest using HttpBadRequestError

apps/io-func-sign-user/src/infra/azure/functions/get-qtsp-clauses-metadata.ts
```diff
@@ -10,28 +10,31 @@ import { error, success } from "@internal/io-sign/infra/http/response";
...
+const decodeHttpRequest = flow(azure.fromHttpRequest, TE.fromEither);
+
+export const makeGetQtspClausesMetadataFunction = (config: NamirialConfig) =>
+  pipe(
+    createHandler(
+      decodeHttpRequest,
+      () =>
+        pipe(
+          getQtspClausesWithToken(config),
+          TE.mapLeft((e) => new HttpBadRequestError(e.message)),
```
gunzip, line 33: @grausof isn't this already handled inside "error" method?

grausof: Nope! Any Error becomes:
```
{
    "title": "Something went wrong.",
    "status": 500
}
```

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -1,96 +1,97 @@
...
+            headers: {
+              "Content-Type": "application/json",
+            },
+          }),
+        E.toError
+      ),
+      TE.filterOrElse(
+        is2xx,
+        () => new Error("The attempt to get Namirial token failed.")
+      ),
```
grausof, line 41: this code with the addition of new endpoints will often be replicated. How about writing a function that takes as input only the error message and the `Request`?

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -1,96 +1,97 @@
...
+      TE.of(token),
+      TE.chain((token) =>
+        TE.tryCatch(
+          () =>
+            fetchWithTimeout(`${basePath}/api/tos/`, {
+              method: "GET",
+              headers: {
+                "Content-Type": "application/json",
+                Authorization: `Bearer ${token.access}`,
+              },
```
grausof, line 70: Here too this part will be repeated often.

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -1,96 +1,97 @@
...
-      },
-    })
-      .then(
-        responseToJson(
-          "The attempt to retrieve the the auth token from the QTSP failed."
+      ),
+      TE.filterOrElse(
+        is2xx,
+        () => new Error("The attempt to get Namirial clauses failed.")
+      ),
```
grausof, line 78: ditto

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -1,96 +1,97 @@
...
+      TE.chain((response) => TE.tryCatch(() => response.json(), E.toError)),
+      TE.chainEitherKW(
+        flow(
+          ClausesMetadata.decode,
+          E.mapLeft(
+            (errs) =>
+              new Error(
+                `Invalid format for Namirial clauses: ${readableReport(errs)}`
+              )
+          )
```
grausof, line 88: Here too this part will be repeated often. How about replacing this code with a function

gunzip: I'd do that when and in case it will be useful. Now it's not because there are some moving parts (error message, filter condition, type of decoding...) that would make the isolation of this logic useless atm.

grausof (approved):

grausof:
@gunzip LGTM! If there are no other small refactors you can proceed with the merge and we continue on #35

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -1,96 +1,97 @@
...
 
-const httpApiFetch = agent.getHttpFetch(process.env);
-const abortableFetch = AbortableFetch(httpApiFetch);
-
-const getAuthHeaders = (accessToken: NonEmptyString) => ({
-  "Content-Type": "application/json",
-  Authorization: `Bearer ${accessToken}`,
-});
-
 const is2xx = (r: Response): boolean => r.status >= 200 && r.status < 300;
```
lucacavallaro, line 18: Only a minor thing: cab we name it "isSuccessful"? 

The "is2xx" name leaks details about the implementation, it is a bit "imperative"

lucacavallaro (approved):

PR #37, opened 2022-12-02
chore(user): [SFEQS-1138] Minor fix openapi

grausof:
Minor fix to openapi

### DIFF
--- a/apps/io-func-sign-user/openapi.yaml
+++ b/apps/io-func-sign-user/openapi.yaml
@@ -97,10 +97,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/QtspClausesMetadataDetailView"
-        "403":
-          $ref: "#/components/responses/Forbidden"
-        "404":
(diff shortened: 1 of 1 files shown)


gunzip (approved):
can you link the jira issue?

lucacavallaro:
> can you link the jira issue?

Is the JIRA issue needed for this kind of changes? we can use this PR as reference

grausof:
> can you link the jira issue?

Yes, this is a minor fix to #35 
I have however updated the title with the main jira id

gunzip:
@lucacavallaro  what do you mean with "we can use this PR as reference"? we have a "contract" in swarmia set up that count unlinked PRs. If we want to have unlinked PR I'd rather switch off this control.

lucacavallaro:
> @lucacavallaro what do you mean with "we can use this PR as reference"? we have a "contract" in swarmia set up that count unlinked PRs. If we want to have unlinked PR I'd rather switch off this control.

No ok, you are right. We can keep the contract!

PR #43, opened 2022-12-06
feat(user): add createSignature endpoint [SFEQS-1208]

grausof:
This PR adds the signature creation endpoint (`createSignature`). It also brings new implementations on the `Namirial` client. 
It is still a draft as it depends on [SFEQS-1127] and also lacks the response encoder. 
Polling to the qtsp is also missing.

#### List of Changes
- Add `createSignature` endpoint
- Add `createSignatureRequest` to Namirial client

Locally tested

#### Checklist:


apps/io-func-sign-user/openapi.yaml
```diff
@@ -301,3 +332,144 @@ components:
...
+        - clauses
+        - document_url
+        - privacy_url
+        - privacy_text
+        - terms_and_conditions_url
+        - nonce
+
+    PublicKey:
+      type: string
+      format: base64url
```
gunzip, line 345: are you sure this is right? I haven't found any doc about `base64url` format

apps/io-func-sign-user/src/infra/http/decoder/document-to-sign.ts
```diff
@@ -0,0 +1,95 @@
...
+  DocumentToSignToApiModel.encode
+);
+
+export const requireDocumentsSignature = flow(
+  (res: HttpRequest) => res.body,
+  validate(CreateSignatureBody),
+  E.map((body) => body.documents_to_sign),
+  E.chain(
+    validate(
+      tx.nonEmptyArray(DocumentToSignApiModel.pipe(DocumentToSignFromApiModel)),
```
gunzip, line 91: consider using `NonEmptyArray` from fp-ts so you can avoid to import io-ts-types

apps/io-func-sign-user/openapi.yaml
grausof: I've found a reference on [Zalando guidelines](http://opensource.zalando.com/restful-api-guidelines/#data-formats) which suggests to use `base64url` for base64 content. However this reference (`PublicKey`) will not be used for now so I will have to delete it.

apps/io-func-sign-user/src/infra/http/decoder/document-to-sign.ts
grausof: I can't use `NonEmptyArray` in this case because I need the type while in `fp-ts` it's a value.

apps/io-func-sign-user/openapi.yaml
gunzip: I think they mean to use format: binary with contentEncoding: base64url

apps/io-func-sign-user/src/app/main.ts
```diff
@@ -32,12 +35,17 @@ const database = cosmosClient.database(config.azure.cosmos.dbName);
...
 
 const documentsToFillQueue = new QueueClient(
   config.azure.storage.connectionString,
-  config.documentsToFillQueueName
+  "waiting-for-documents-to-fill"
+);
+
+const qtspQueue = new QueueClient(
+  config.azure.storage.connectionString,
+  "waiting-for-qtsp"
```
lucacavallaro, line 48: what kind of messages are enqueued to `waiting-for-qtsp` queue? Is this storage queue shared between different workflows?

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -32,7 +33,7 @@ export type CreateFilledDocumentPayload = t.TypeOf<
  */
 export const makeCreateFilledDocumentUrl =
   (
-    getFilledDocumentUrl: GetBlobUrl,
+    getSasFilledDocumentUrl: GetSasFilledDocumentUrl,
```
lucacavallaro, line 36: the `application layer` (`./app` folder) should not be aware of the implementation details of the infrastructure

`Sas` is not a `domain` entity/concept

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
```diff
@@ -32,7 +33,7 @@ export type CreateFilledDocumentPayload = t.TypeOf<
...
     enqueueDocumentToFill: EnqueueMessage,
```
lucacavallaro, line 37: same here: `EnqueueMessage` name leaks the information that we are using a `queue`, the `use-case` can't be coupled to this implementation detail 

maybe for this release we can keep these names, but should be changed ASAP

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
```diff
@@ -28,20 +36,39 @@ import { CreateFilledDocumentBody } from "../../http/models/CreateFilledDocument
...
+  ) =>
+    pipe(
+      filledDocumentBlobName,
+      getBlobClient,
+      RTE.chainTaskEitherK(
+        generateSasUrlFromBlob(
+          pipe(
+            defaultBlobGenerateSasUrlOptions(),
+            withPermissions("r"),
+            withExpireInMinutes(120)
```
lucacavallaro, line 64: Why `120` minutes?

apps/io-func-sign-user/src/infra/http/decoder/qtsp-clause.ts
```diff
@@ -0,0 +1,49 @@
...
+  validate(CreateSignatureBody),
+  E.map((body) => body.qtsp_clauses),
+  (qtspClauses) =>
+    sequenceS(E.Apply)({
+      acceptedClauses: pipe(
+        qtspClauses,
+        E.map((qtspClauses) => qtspClauses.accepted_clauses),
+        E.chain(
+          validate(
+            tx.nonEmptyArray(QtspClauseApiModel.pipe(QtspClauseFromApiModel)),
```
lucacavallaro, line 37: `tx.nonEmptyArray(QtspClauseApiModel.pipe(QtspClauseFromApiModel)` it's difficult to read, can you extract it in a separate decoder?

apps/io-func-sign-user/src/infra/http/encoders/namirial-clauses-metadata.ts
```diff
@@ -1,6 +1,6 @@
 import * as E from "io-ts/lib/Encoder";
-import { QtspClausesMetadata } from "../../../qtsp-clauses-metadata";
-import { ClausesMetadata } from "../../namirial/clauses-metadata";
+import { QtspClausesMetadata } from "../../../qtsp";
+import { ClausesMetadata } from "../../namirial/types/clauses-metadata";
```
lucacavallaro, line 3: avoid using "types" folder

1. It's not the way we're using to group modules in our project
2. It's not idiomatic in JavaScript/TypeScript, libraries such as `fp-ts` doesn't expose `types` with import path like `fp-ts/lib/Either/types`

apps/io-func-sign-user/src/infra/namirial/client.ts
```diff
@@ -31,6 +36,7 @@ export const makeGetToken =
             }),
             method: "POST",
             headers: {
+              // eslint-disable-next-line sonarjs/no-duplicate-string
```
lucacavallaro, line 39: maybe we can extract `headers` (and other common properties) in a separate variable? we can make a "builder" function that fills this common properties

apps/io-func-sign-user/src/infra/namirial/types/signature-request.ts
```diff
@@ -0,0 +1,81 @@
...
+  nonce: NonEmptyString,
+  tos_signature: NonEmptyString,
+  signatures: Signature,
+});
+
+export type CreateSignatureRequestBody = t.TypeOf<
+  typeof CreateSignatureRequestBody
+>;
+
+enum SignatureRequestStatus {
```
lucacavallaro, line 52: avoid using `enum`, prefer `union types` with literal strings

`type SignatureRequestStatus = "CREATED" | "READY" | ...`

apps/io-func-sign-user/src/signature.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as TE from "fp-ts/lib/TaskEither";
+import * as O from "fp-ts/lib/Option";
+
+import { IsoDateFromString } from "@pagopa/ts-commons/lib/dates";
+
+import { Id, id as newId } from "@io-sign/io-sign/id";
+
+import { Signer } from "@io-sign/io-sign/signer";
+
+export enum SignatureStatus {
```
lucacavallaro, line 12: ditto (avoid using `enum`)

apps/io-func-sign-user/src/signature.ts
```diff
@@ -0,0 +1,69 @@
...
+  CREATED = "CREATED",
+  READY = "READY",
+  WAITING = "WAITING",
+  COMPLETED = "COMPLETED",
+  FAILED = "FAILED",
+}
+
+const SignatureStatusV = t.keyof({
```
lucacavallaro, line 20: `t.keyof` builds a `sum type`, so I don't understand why you created `SignatureStatus` as `enum`

----

`V` in `SignatureStatusV` stands for?

lucacavallaro (changes requested):

apps/io-func-sign-user/src/app/main.ts
grausof: Nope, this storage queue is for polling the QTSP

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
grausof: ok, I'll rename it!

apps/io-func-sign-user/src/app/use-cases/create-filled-document.ts
grausof: You are right! U will try to make a reafctor of this part.

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
grausof: This URL is passed to the `QTSP` which then verifies it and issues the signature with relative certificate. The QTSP may also take a long time to process a signature, so I think `120 minutes` is reasonable.

lucacavallaro: maybe in the future we can proxy the URL so that we can generate smaller tokens

(it's ok 120 for now)

lucacavallaro (approved):

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
```diff
@@ -0,0 +1,183 @@
...
+      );
+    return pipe(
+      signerId,
+      getSignature(signatureId),
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError(`Signature ${signatureId} not found`)
+        )
+      ),
+      TE.chainW((signature) =>
```
lucacavallaro, line 93: I think that this flow is too complex, it should be refactored to make it more "linear" and declarative

I'll add a JIRA card for refactor

lucacavallaro (approved):

PR #45, opened 2022-12-07
feat: [SFEQS-1204, SFEQS-1214] Implement functions: CreateSignatureRequest, GetSignatureRequest, MarkAsWaitForSignature

lucacavallaro:
#### List of Changes

In `@io-sign/io-sign`:
1. Implemented new common structures (`SignatureRequest`, `Issuer`)
2. Add utilities to work with Azure blob and queues

In `io-func-sign-user`:
1. Implement `CreateSignatureRequest`, `GetSignatureRequest` Azure Functions
4. Implement `Signature Request` entity
5. Add `SAS Url` generation to make `documents` readable from mobile app
6. Add an event emitter (backed by a `Storage Queue`) to notify the changed status of a Signature Request
7. Add a function to convert `io-func-sign-issuer`'s `Signature Field` attributes to the user's one

In `io-func-sign-issuer`:
1. Implement `MarkAsWaitForSignature` Azure Function to update the `Signature Request` status when persisted on `io-func-sign-user` backend
2. Refactor `Signature Request` to enhance type safety (now is a sum type)

It was e2e tested manually in local and development (remote) environment.

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
+++ b/apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
@@ -1,19 +1,24 @@
 import { pipe } from "fp-ts/lib/function";
 import * as TE from "fp-ts/lib/TaskEither";
 import * as t from "io-ts";
+
 import { GetFiscalCodeBySignerId } from "@io-sign/io-sign/signer";
-import { Notification } from "@io-sign/io-sign/notification";
+
 import {
   SubmitMessageForUser,
   withFiscalCode,
 } from "@io-sign/io-sign/infra/io-services/message";
+
 import { validate } from "@io-sign/io-sign/validation";
 import { sequenceS } from "fp-ts/lib/Apply";
 import { EntityNotFoundError } from "@io-sign/io-sign/error";
+
+import { SignatureRequestToBeSigned } from "@io-sign/io-sign/signature-request";
 import {
   SignatureRequest,
   UpsertSignatureRequest,
 } from "../../signature-request";
+
 import { Dossier, GetDossier } from "../../dossier";
 
 export type SendNotificationPayload = {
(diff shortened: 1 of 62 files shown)


apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
```diff
@@ -52,53 +57,54 @@ const makeMessage =
...
+          ),
+          TE.fromEither
+        ),
+        notification: pipe(
+          getFiscalCodeBySignerId(signatureRequest.signerId),
+          TE.chain(
+            TE.fromOption(
+              () =>
+                new EntityNotFoundError(
+                  "The tax code associated with this signer is not valid!"
```
grausof, line 92:
```suggestion
                  "The fiscal code associated with this signer is not valid!"
```

apps/io-func-sign-user/src/infra/azure/functions/get-signature-request.ts
```diff
@@ -0,0 +1,92 @@
...
+    document.url,
+    split("/"),
+    last,
+    getBlobClient,
+    RTE.chainTaskEitherK(
+      generateSasUrlFromBlob(
+        pipe(
+          defaultBlobGenerateSasUrlOptions(),
+          withPermissions("r"),
+          withExpireInMinutes(5)
```
grausof, line 46: Is 5 enough?

apps/io-func-sign-user/src/signature-field.ts
```diff
@@ -0,0 +1,47 @@
...
+  size,
+  page,
+}: AttributesWithCoordsAndSize): SignatureFieldToBeCreatedAttributes => ({
+  bottomLeft: coordinates,
+  topRight: {
+    x: coordinates.x + size.w.valueOf(),
+    y: coordinates.y + size.h.valueOf(),
+  },
+  page,
+});
```
grausof, line 37: Here it is also necessary to change the reference system. On the issuer side the zero point is at the top left while on the user side it is at the bottom left. You need to read the document via pdf-handler and calculate the page height. We can also do this with a separate PR but we need to enter a TODO

apps/io-func-sign-user/src/infra/http/encoders/signature-field.ts
```diff
@@ -0,0 +1,67 @@
...
+      ({ uniqueName: unique_name }) => ({
+        unique_name,
+      }),
+      flow(
+        toSignatureFieldToBeCreatedAttributes,
+        ({ bottomLeft: bottom_left, topRight: top_right, page }) => ({
+          bottom_left,
+          top_right,
+          page,
+        })
```
grausof, line 63: I would move this part before saving the signature request on the DB in such a way as to save it already with the modified coordinates. Based on the previous comment the calculation may take some time.

grausof (changes requested):

apps/io-func-sign-user/src/signature-field.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const toSignatureFieldToBeCreatedAttributes = ({
+  coordinates,
+  size,
+  page,
+}: AttributesWithCoordsAndSize): SignatureFieldToBeCreatedAttributes => ({
+  bottomLeft: coordinates,
+  topRight: {
+    x: coordinates.x + size.w.valueOf(),
+    y: coordinates.y + size.h.valueOf(),
```
grausof, line 34: `+` causes the code review pipeline to fail.

apps/io-func-sign-user/src/signature-field.ts
lucacavallaro: Ok, I totally missed this! 

I'll split the task again. And you are right: if there is an async computation we have to store the result in the DB. 

Maybe we can store this kind of information (for example the document height) at "ValidateDocument" step.

apps/io-func-sign-user/src/infra/azure/functions/get-signature-request.ts
lucacavallaro: I think 5 minutes is enough, if not too much.

If you think about it, the app downloads PDFs immediately after making this HTTP call. It could also be set to 1 minute.

apps/io-func-sign-user/extensions.csproj
```diff
@@ -0,0 +1,11 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+	<WarningsAsErrors></WarningsAsErrors>
+	<DefaultItemExcludes>**</DefaultItemExcludes>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.Storage" Version="5.0.1" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Script.ExtensionsMetadataGenerator" Version="1.1.3" />
+  </ItemGroup>
+</Project>
```
grausof, line 11: Is there a need to commit it?

apps/io-func-sign-user/openapi.yaml
```diff
@@ -314,3 +346,139 @@ components:
...
+          type: string
+          format: uri
+      required:
+        - id
+        - status
+        - signer_id
+        - expires_at
+        - documents
+        - created_at
+        - updated_at
```
grausof, line 484:
```suggestion
        - id
        - status
        - dossier_id
        - signer_id
        - expires_at
        - documents
        - created_at
        - updated_at
```

apps/io-func-sign-user/extensions.csproj
lucacavallaro: yes

grausof (approved):
LGTM 🎉

PR #52, opened 2022-12-14
chore(issuer): [SFEQS-1224] Align notification message

grausof:
This PR aligns the signature message with what is intended by design


lucacavallaro (approved):
Can you add a "patch" changeset file?

PR #54, opened 2022-12-15
chore(user): Mock crypto signatures [SFEQS-1210]

grausof:
This PR mocks Lollipop for signing the content to be sent to the QTSP.

#### List of Changes
- Add `crypto-js` and `jsrsasign` library
- Create a valid signature for `tosSignature` and `signature` field
- Add workaround for `document_url` and change openapi format to base64
- Fix success and error message mock
- Fix a bug on `getSignatureRequest`  which always returned the validated url and not the signed one.

Tested in local with dev env

#### Checklist:


apps/io-func-sign-user/src/infra/azure/functions/create-signature.ts
```diff
@@ -104,7 +105,26 @@ const makeCreateSignatureHandler = (
         body: { email, signatureRequestId },
       }) => ({
         signer,
-        qtspClauses,
+        qtspClauses: {
+          ...qtspClauses,
+          // TODO: [SFEQS-1237] workaround for WAF
```
gunzip, line 110: I guess this requires a change to the io-app code as well, isn't it?

apps/io-func-sign-user/ValidateQtspSignature/function.json
```diff
@@ -13,7 +13,7 @@
   "entryPoint": "ValidateSignature",
   "retry": {
         "strategy": "fixedDelay",
-        "maxRetryCount": 1,
-        "delayInterval": "00:00:10"
+        "maxRetryCount": 2,
+        "delayInterval": "00:01:00"
```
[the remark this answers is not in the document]
grausof: Yes, the QTSP takes about 20/30 seconds now that it's unloaded. So I think 1 minute is enough.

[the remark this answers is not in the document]
grausof: consider that however the QTSP is in UAT! maybe I'll turn it down to 30 seconds.

apps/io-func-sign-user/src/infra/azure/functions/create-signature.ts
lucacavallaro: @gunzip yes

grausof: Only the type in the openapi needs to be changed. But since url or string is always mapped as `NonEmptyString` by generator there is no problem.

lucacavallaro (approved):

PR #55, opened 2022-12-15
fix: [SFEQS-1238] Round coordinates for QTSP 

lucacavallaro:
#### List of Changes

1. Round coordinates in `io-func-sign-user`, now are all integers

#### Motivation and Context

The QTSP accepts only integer coordinates

#### Checklist:

### DIFF
--- a/apps/io-func-sign-user/src/signature-field.ts
+++ b/apps/io-func-sign-user/src/signature-field.ts
@@ -32,7 +32,7 @@ export const toSignatureFieldToBeCreatedAttributes =
   }: AttributesWithCoordsAndSize): SignatureFieldToBeCreatedAttributes => ({
     bottomLeft: {
       x: Math.round(coordinates.x),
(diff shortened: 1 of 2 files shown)


apps/io-func-sign-user/src/signature-field.ts
```diff
@@ -31,12 +31,12 @@ export const toSignatureFieldToBeCreatedAttributes =
     page,
   }: AttributesWithCoordsAndSize): SignatureFieldToBeCreatedAttributes => ({
     bottomLeft: {
-      x: coordinates.x,
-      y: pageHeight - coordinates.y + size.h,
+      x: Math.round(coordinates.x),
+      y: Math.round(pageHeight - coordinates.y + size.h),
```
grausof, line 35:
```suggestion
      y: Math.round(pageHeight - coordinates.y -  size.h),
```
we are always thinking from the opposite point of view!

grausof (approved):

PR #58, opened 2022-12-16
chore: [SFEQS-1250] Add WAIT_FOR_QTSP status to signature request

grausof:
This PR add the `WAIT_FOR_QTSP` status to Signature Request.

### DIFF
--- a/apps/io-func-sign-user/src/signature-request.ts
+++ b/apps/io-func-sign-user/src/signature-request.ts
@@ -2,6 +2,7 @@ import {
   SignatureRequestToBeSigned,
   SignatureRequestRejected,
   SignatureRequestSigned,
+  SignatureRequestWaitForQtsp,
 } from "@io-sign/io-sign/signature-request";
 
 import { Id } from "@io-sign/io-sign/id";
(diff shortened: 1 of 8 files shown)


apps/io-func-sign-user/src/signature-request.ts
```diff
@@ -83,12 +129,13 @@ const onWaitForSignatureStatus =
       default:
         return E.left(
           new ActionNotAllowedError(
-            "This operation is prohibited if the signature request is in WAIT_FOR_SIGNATURE status"
+            "This operation is prohibited if the signature request is in WAIT_FOR_QTSP status"
```
gunzip, line 132: I'd refactor all "This operation" statements with the name of the operation (action/status), otherwise it's hard to tell what's happening from logs.

grausof: I fixed this part also on the issuer side!

packages/io-sign/src/signature-request.ts
```diff
@@ -58,6 +58,24 @@ export type SignatureRequestToBeSigned = t.TypeOf<
   typeof SignatureRequestToBeSigned
 >;
 
+export const SignatureRequestWaitForQtsp = makeSignatureRequestVariant(
+  "WAIT_FOR_QTSP",
+  t.intersection([
+    t.type({
+      qrCodeUrl: t.string,
+      documents: t.array(DocumentReady),
+      sendToQtspAt: IsoDateFromString,
```
lucacavallaro, line 67: I think that this date is not relevant (we can use the `update_at` field for this purpose) so imho should be removed

lucacavallaro (approved):

PR #59, opened 2022-12-21
chore: [SFEQS-1269] Create a custom SAML mock

grausof:
This PR creates a custom Spid Assertion Mock with user data contained in the pdf signature request form.
⚠️ Before deployment requires a change to the `SpidAssertionMock` environment variable

#### Motivation and Context
With this change, the name of the signer on signed PDF will no longer be the same but varies according to the signer.

#### Checklist:

### DIFF
--- a/packages/io-sign/src/infra/pdf.ts
+++ b/packages/io-sign/src/infra/pdf.ts
@@ -66,15 +66,7 @@ export type Field = {
 
 const populate = (form: PDFForm) => (field: Field) =>
   pipe(
-    E.tryCatch(
-      () => form.getTextField(field.fieldName),
-      (e) =>
-        e instanceof Error
-          ? new EntityNotFoundError(e.message)
-          : new EntityNotFoundError(
-              "An error occurred while attempting to access the pdf field."
-            )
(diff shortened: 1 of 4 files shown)


packages/io-sign/src/infra/pdf.ts
```diff
@@ -72,12 +72,42 @@ const populate = (form: PDFForm) => (field: Field) =>
...
+  (form: PDFForm) =>
+  (fieldName: string): E.Either<EntityNotFoundError, Field> =>
+    pipe(
+      E.tryCatch(
+        () => form.getTextField(fieldName),
+        // eslint-disable-next-line sonarjs/no-identical-functions
+        (e) =>
+          e instanceof Error
+            ? new EntityNotFoundError(e.message)
+            : new EntityNotFoundError(
```
lucacavallaro, line 91: after receiving this kind of error, how a consumer can recover?

grausof: Right. I converted it to an internal error.

lucacavallaro (approved):

PR #63, opened 2023-01-11
feat: [SFEQS-1273] Adds the health check on the infrastructure through the info endpoint

grausof:
This PR adds the health check via the info endpoint. In particular, make a check on:

* io-func-sign-user
    * Namirial endpoints
    * Tokenizer endpoint
    * IO Services endpoint
    * Azure Cosmos DB
    * Azure Storage Containers
    * Azure Storage Queues
    
* io-func-sign-issuer
    * Tokenizer endpoint
    * IO Services endpoint
    * Azure Cosmos DB
    * Azure Storage Containers
    * Azure Storage Queues
    

#### Checklist:


apps/io-func-sign-user/src/infra/azure/functions/info.ts
```diff
@@ -1,4 +1,92 @@
...
+  documentsToFillQueue: QueueClient,
+  qtspQueue: QueueClient,
+  onWaitForSignatureQueueClient: QueueClient
+) =>
+  createHandler(
+    nopRequestDecoder,
+    () =>
+      pipe(
+        [
+          makeNamirialHealthCheck(namirialConfig),
```
gunzip, line 73: nit: is this body executed for every request? in case you can build this array outside `createHandler` so make* methods are called once.

gunzip (approved):

apps/io-func-sign-user/src/infra/azure/functions/info.ts
grausof: Each `make` method performs the health check on the endpoints so it must be called on each request. Eventually I can rename the method!

gunzip: Now I get it, I erroneously thought that the make methods where just factories.

PR #64, opened 2023-01-16
fix: [SFEQS-1182, SFEQS-1244, SFEQS-1221] Fix rest API responses

lucacavallaro:
#### List of Changes

1. `rejected_at` is now returned when a `SignatureRequest` is `REJECTED`
2. `CreateDossier` returns `400 (Bad Request)` instead of `500 (Internal Server Error)` when validation fails on `documents_metadata`
3. `CreateSignatureRequest` returns `400 (Bad Request)` instead of `500 (Internal Server Error)` when validations fails on `expires_at`

#### Checklist:

### DIFF
--- a/.changeset/lovely-days-admire.md
+++ b/.changeset/lovely-days-admire.md
@@ -1,0 +1,6 @@
+---
+"io-func-sign-issuer": patch
+"@io-sign/io-sign": patch
+---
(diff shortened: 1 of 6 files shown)


grausof (approved):
LGTM

PR #66, opened 2023-01-17
feat!: [SFEQS-1266, SFEQS-1216] Add signatureFields validation with document metadata

grausof:
⚠ **BREAKING CHANGE:** Added `formFields` attribute to `DocumentMetadata` in `io-sign` package!
(Previously generated signing requests are not compatible with this version!)

This PR add the `signatureFields` validation with document metadata, in detail, carry out the following checks:

- Check that the uploaded file is a PDF
- Check that in the case of _signature fields_, they are present in the PDF (check on the `fieldId`)
- Check the type of signature field that is compatible with the _PADeS_ signature
- Check that in case of coordinates the page number exists inside the PDF
- Check that in the case of coordinates, the box is inside the pdf page.

Furthermore, this PR solves a TODO [SFEQS-1216] by removing the dependencies from `infra` within the use case.

#### Checklist:

[SFEQS-1216]: https://pagopa.atlassian.net/browse/SFEQS-1216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -12,14 +20,77 @@ import {
...
+  (
+    pages: DocumentMetadata["pages"],
+    formFields: DocumentMetadata["formFields"]
+  ) =>
+  (signatureFields: DocumentMetadata["signatureFields"]) =>
+    pipe(
+      signatureFields,
+      A.map((signatureField) => signatureField.attributes),
+      A.map((attributes) =>
+        SignatureFieldAttributes.is(attributes)
```
gunzip, line 49: Is it's possibile I would break this method into smaller ones to increase readability (you can better graps what's happening when using method names)

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -12,14 +20,77 @@ import {
...
+                  : E.left(
+                      new Error(
+                        `The dossier signature field (${attributes.uniqueName}) doesn't appear to be a signature field.`
+                      )
+                    )
+              )
+            )
+          : pipe(
+              pages,
+              A.filter((p) => p.number === attributes.page),
```
gunzip, line 72: You may consider to alias these conditions (ie. `hasSomething` / `isSomething`) as well

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -12,14 +20,77 @@ import {
...
+              A.head,
+              E.fromOption(
+                () =>
+                  new Error(
+                    `The page number in the dossier signature field (${attributes.page}) was not found in the uploaded document.`
+                  )
+              ),
+              E.chain((page) =>
+                attributes.coordinates.x + attributes.size.w < page.width &&
```
gunzip, line 81: ditto

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -75,31 +147,62 @@ export const makeValidateUpload =
...
+                    "The specified document does not exists."
+                  )
+              ),
+              E.chain((document) =>
+                pipe(
+                  document.metadata.signatureFields,
+                  validateSignatureFieldsWithMetadata(pages, formFields),
+                  A.separate,
+                  (validationResults) =>
+                    pipe(validationResults.left, A.isEmpty)
```
gunzip, line 166: It's hard to understand what's happening here, same considerations as above (or add a comment)

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -75,31 +147,62 @@ export const makeValidateUpload =
...
-          TE.chainEitherK(({ metadata: { pages }, url }) =>
+          TE.chainEitherK(({ metadata: { pages, formFields }, url }) =>
             pipe(
               signatureRequest,
-              markDocumentAsReady(uploadMetadata.documentId, url, pages)
+              getDocument(uploadMetadata.documentId),
+              E.fromOption(
+                () =>
+                  new EntityNotFoundError(
+                    "The specified document does not exists."
```
gunzip, line 157:
```suggestion
                    "No document was found with the specified document id."
```

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -75,31 +147,62 @@ export const makeValidateUpload =
...
-                "There was an error on the upload"
+                "The uploaded file appears to be corrupted or does not appear to be a PDF file."
               ),
               TE.fromEither
             )
           )
         )
       ),
+      TE.chainFirst((signatureRequest) =>
+        upsertUploadMetadata({
```
gunzip, line 200: I'd alias this method as well (ie. `markAsReady`)

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
gunzip, line 166: lacks unit tests

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -12,14 +20,77 @@ import {
...
+              E.fromOption(
+                () =>
+                  new Error(
+                    `The page number in the dossier signature field (${attributes.page}) was not found in the uploaded document.`
+                  )
+              ),
+              E.chain((page) =>
+                attributes.coordinates.x + attributes.size.w < page.width &&
+                attributes.coordinates.y + attributes.size.h < page.height
+                  ? E.right(true)
```
gunzip, line 83: add unit tests

gunzip (changes requested):

packages/io-sign/src/document.ts
```diff
@@ -68,6 +69,11 @@ export const SignatureField = t.type({
 
 export type SignatureField = t.TypeOf<typeof SignatureField>;
 
+export enum FormFieldTypeEnum {
+  "SIGNATURE" = "SIGNATURE",
+  "OTHER" = "OTHER",
+}
```
lucacavallaro, line 75: 1. use a `union type` instead of an `enum`
2. avoid using "suffix" such as `Enum`, put directly the name of the type (`FormFieldType`)

```suggestion
export type FormFieldType = "SIGNATURE" | "OTHER"; 
```

moreover, if the only value allowed in our case is "SIGNATURE" make sure that a different value is never accepted at the outset (traces "OTHER" only if we do something with it later)

packages/io-sign/src/document.ts
```diff
@@ -230,7 +247,10 @@ const onWaitForValidationStatus =
     switch (action.name) {
       case "MARK_AS_READY":
         return E.right(
-          updateMetadataPage(action.payload.pages)({
+          updateMetadataPageAndFields(
```
lucacavallaro, line 250: if `pages` and `formFields` must always be updated together, try to extract them as a single `value object` (`PDFDocument`)

packages/io-sign/src/infra/pdf.ts
```diff
@@ -22,6 +22,12 @@ export const PdfMetadata = t.intersection([
         height: NonNegativeNumber,
       })
     ),
+    fields: t.array(
```
lucacavallaro, line 25: I think this module can be treated like an "adapter", maybe we can reuse the types defined on the domain part

packages/io-sign/src/__test__/document.spec.ts
```diff
@@ -82,7 +83,7 @@ describe("Document", () => {
       expect(
         pipe(
           rejectedDocument,
-          E.chain(markAsReady("https://my.document.url", [])),
+          E.chain(markAsReady("https://my.document.url", [], [])),
```
lucacavallaro, line 86: I think that (`[], []`) is a bit ugly and confusing, since they share their context and lifecycle they can be composed to form a singular value object

lucacavallaro (changes requested):

packages/io-sign/src/document.ts
```diff
@@ -78,6 +84,12 @@ export const DocumentMetadata = t.type({
       height: NonNegativeNumber,
     })
   ),
+  formFields: t.array(
```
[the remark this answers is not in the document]
grausof: This information has the same value as pages. I will try to merge both information into one attribute!

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -1,25 +1,117 @@
...
+export type GetPdfMetadata = (
+  buffer: Buffer
+) => TE.TaskEither<Error, PdfDocumentMetadata>;
+
+const getPage =
+  (pageNumber: NonNegativeNumber) => (pages: PdfDocumentMetadata["pages"]) =>
+    pipe(
+      pages,
+      A.filter((p) => p.number === pageNumber),
+      A.head
```
gunzip, line 44: you can use `findFirst`

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -1,25 +1,117 @@
...
+  (page: PdfDocumentMetadataPage) =>
+    fieldAttributes.coordinates.x + fieldAttributes.size.w < page.width &&
+    fieldAttributes.coordinates.y + fieldAttributes.size.h < page.height;
+
+const validateSignatureField =
+  (formFields: PdfDocumentMetadata["formFields"]) =>
+  (attributes: SignatureFieldAttributes) =>
+    pipe(
+      formFields,
+      A.filter((field) => field.name === attributes.uniqueName),
```
gunzip, line 58: ditto

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -1,25 +1,117 @@
...
+      A.head
+    );
+
+const isFieldInsidePage =
+  (fieldAttributes: SignatureFieldToBeCreatedAttributes) =>
+  (page: PdfDocumentMetadataPage) =>
+    fieldAttributes.coordinates.x + fieldAttributes.size.w < page.width &&
+    fieldAttributes.coordinates.y + fieldAttributes.size.h < page.height;
+
+const validateSignatureField =
```
gunzip, line 53:
```suggestion
const isValidSignatureField =
```

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -1,25 +1,117 @@
...
+      A.head,
+      E.fromOption(() => [
+        new Error(
+          `The dossier signature field (${attributes.uniqueName}) was not found in the uploaded document.`
+        ),
+      ]),
+      E.map(() => true)
+    );
+
+const validateSignatureFieldToBeCreated =
```
gunzip, line 68:
```suggestion
const isValidaSignatureFieldToBeCreated =
```

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -1,25 +1,117 @@
...
+    pipe(
+      pages,
+      getPage(attributes.page),
+      E.fromOption(() => [
+        new Error(
+          `The page number in the dossier signature field (${attributes.page}) was not found in the uploaded document.`
+        ),
+      ]),
+      E.map(isFieldInsidePage(attributes)),
+      E.chain((isValid) =>
```
gunzip, line 80: you can use `E.mapLeft` instead

gunzip (commented):
better 👍

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
grausof: `isValid` is a boolean and if it is `false` I have to return a `leftEither`. With `mapLeft` instead I lose control over `isValid`.

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
grausof: I'm having trouble with unit tests. They run locally but the code review pipeline fails!

gunzip: can you share a link to the action where they fail?

grausof: sure! https://github.com/pagopa/io-sign/actions/runs/3943288342

grausof: I probably found the cause, by directly launching` yarn run code-review` the pipeline fails, instead if a `yarn workspaces foreach run test` is first launched then the code-review is successful! Sounds like a turbo issue!

gunzip: using "turbo run typecheck lint test" locally ?

grausof: i get the same pipeline error! If I then run `yarn workspaces foreach run test`, `turbo run typecheck lint test` works

grausof: This is the error I get when I run the test locally the first time. The strange thing is that if I rerun the test no errors are returned. To reproduce it I had to use the `yarn jest --clearCache` command!

```
src/app/use-cases/validate-upload.ts:108:7 - error TS2345: Argument of type '<A>(ta: HKT<"Either", A>[]) => HKT<"Either", A[]>' is not assignable to parameter of type '(c: Either<string, boolean>[]) => HKT<"Either", unknown[]>'.
      Types of parameters 'ta' and 'c' are incompatible.
        Type 'Either<string, boolean>[]' is not assignable to type 'HKT<"Either", unknown>[]'.
          Type 'Either<string, boolean>' is not assignable to type 'HKT<"Either", unknown>'.
            Type 'Left<string>' is missing the following properties from type 'HKT<"Either", unknown>': _URI, _A

    108       A.sequence(applicativeValidation)
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    src/app/use-cases/validate-upload.ts:138:9 - error TS2322: Type 'unknown' is not assignable to type 'TaskEither<Error, unknown>'.

    138         pipe(
                ~~~~~
    139           sequenceS(TE.ApplySeq)({
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    ... 
    213           )
        ~~~~~~~~~~~
    214         )
        ~~~~~~~~~
```

grausof: [This](https://github.com/devexperts/remote-data-ts/issues/49) looks like a similar problem but I can't figure it out! @gunzip

grausof: I replaced `jest` with `vitest` which fixed the problem as it looks like a `jest` build problem that only happens in this particular repo setup.

gunzip: @lucacavallaro do you see any issue using vitest instead of jest? (we must add an ADR in case pointing to this PR) of course with another PR

grausof: Please check #67

lucacavallaro (approved):

gunzip (approved):

PR #68, opened 2023-01-23
[SFEQS-1222] Keep Signature Request statuses consistent across the system

lucacavallaro:
#### List of Changes

1. Add two functions on the issuer side `MarkAsSigned` and `MarkAsRejected`
2. Add event emitters (backed by Azure Storage Queue)

#### Motivation and Context

since the signature request is stored in two separate databases, we have to keep them in sync

#### Checklist:

- [x] My change requires a change to the documentation.

### DIFF
--- a/apps/io-func-sign-issuer/src/app/use-cases/mark-request-rejected.ts
+++ b/apps/io-func-sign-issuer/src/app/use-cases/mark-request-rejected.ts
@@ -1,0 +1,28 @@
+import { EntityNotFoundError } from "@io-sign/io-sign/error";
+import { SignatureRequestRejected } from "@io-sign/io-sign/signature-request";
+import { pipe } from "fp-ts/lib/function";
+
+import * as TE from "fp-ts/lib/TaskEither";
+
+import {
+  UpsertSignatureRequest,
+  markAsRejected,
+  GetSignatureRequest,
+} from "../../signature-request";
+
+export const makeMarkRequestAsRejected =
+  (
+    getSignatureRequest: GetSignatureRequest,
+    upsertSignatureRequest: UpsertSignatureRequest
+  ) =>
+  (request: SignatureRequestRejected) =>
+    pipe(
+      pipe(request.issuerId, getSignatureRequest(request.id)),
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError("Signature Request not found.")
+        )
+      ),
+      TE.chainEitherK(markAsRejected(request.rejectedAt, request.rejectReason)),
+      TE.chain(upsertSignatureRequest)
+    );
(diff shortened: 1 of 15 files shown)


apps/io-func-sign-issuer/src/app/use-cases/mark-request-rejected.ts
```diff
@@ -0,0 +1,28 @@
...
+  ) =>
+  (request: SignatureRequestRejected) =>
+    pipe(
+      pipe(request.issuerId, getSignatureRequest(request.id)),
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError("Signature Request not found.")
+        )
+      ),
+      TE.chainEitherK(markAsRejected(request.rejectedAt, request.rejectReason)),
```
gunzip, line 26: what if the enque operation fails?

apps/io-func-sign-issuer/src/app/use-cases/mark-request-signed.ts
```diff
@@ -0,0 +1,28 @@
...
+  ) =>
+  (request: SignatureRequestSigned) =>
+    pipe(
+      pipe(request.issuerId, getSignatureRequest(request.id)),
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError("Signature Request not found.")
+        )
+      ),
+      TE.chainEitherK(markAsSigned),
```
gunzip, line 26: ditto

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
```diff
@@ -55,6 +62,9 @@ export const makeMarkSignatureAndSignatureRequestAsRejected =
           markAsRejected(rejectedReason),
           TE.fromEither,
           TE.chain(upsertSignatureRequest),
+          TE.chainFirst((r) =>
```
gunzip, line 65: needs unit tests

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
```diff
@@ -133,11 +146,17 @@ export const makeValidateSignature =
...
                   ),
                   A.sequence(TE.ApplicativeSeq),
+
                   TE.map((documents) => ({
                     ...signatureRequest,
                     documents,
                   })),
                   TE.chainEitherK(markAsSigned),
+                  TE.chainFirst((r: SignatureRequest) =>
+                    notifySignatureRequestSignedEvent(
```
gunzip, line 156: ditto

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
lucacavallaro: this use case is part of a big refactor that I'm developing in parallel and will be released with one of the next PRs

it has been rewritten to make it more linear and have less responsibility, so unit tests will come with that change

apps/io-func-sign-issuer/src/app/use-cases/mark-request-rejected.ts
lucacavallaro: if an error occurs, the signature request is not updated, remains in `WAIT_FOR_QTSP` and is retried later

gunzip (approved):

PR #69, opened 2023-01-24
chore!: [SFEQS-1333] Add the ability to mark an issuer as a no-billable test account

grausof:
This PR add isTrying field to issuer data and remove unused fields.

#### Motivation and Context
The need is to know if an issuer is in the experimental phase or not. We will use this field to communicate to data-lake whether the signing event should be billed or not and at the same time to route the request to the QTSP towards a demo or production environment. The addition has been made now to avoid breaking changes during the initial experimentation.

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
+++ b/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
@@ -15,7 +15,7 @@ const issuer: Issuer = {
   subscriptionId: newId(),
   email: "info@enpacl-pec.it" as EmailString,
   description: "descrizione dell'ente" as NonEmptyString,
-  isTrying: true,
+  isTesting: true,
 };
 
 const dossier = newDossier(issuer, "My dossier", [
--- a/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
+++ b/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
@@ -14,10 +14,8 @@ const issuer: Issuer = {
   id: newId(),
   subscriptionId: newId(),
   email: "info@enpacl-pec.it" as EmailString,
-  address: "Viale Del Caravaggio, 78 - 00147 Roma (RM)",
   description: "descrizione dell'ente" as NonEmptyString,
-  taxCode: "80119170589",
-  vatNumber: "80119170589",
+  isTrying: true,
 };
 
 const dossier = newDossier(issuer, "My dossier", [
--- a/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
+++ b/apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
@@ -15,7 +15,7 @@ const issuer: Issuer = {
   subscriptionId: newId(),
   email: "info@enpacl-pec.it" as EmailString,
   description: "descrizione dell'ente" as NonEmptyString,
-  isTesting: true,
+  environment: "TEST",
 };
 
 const dossier = newDossier(issuer, "My dossier", [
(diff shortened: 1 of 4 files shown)


lucacavallaro:
What is the purpose of this field?

grausof:
> What is the purpose of this field?

The need is to know if an issuer is in the experimental phase or not. We will use this field to communicate to data-lake whether the signing event should be billed or not and at the same time to route the request to the QTSP towards a demo or production environment. The addition has been made now to avoid breaking changes during the initial experimentation.

gunzip:
can you add the reason to the PR / changeset description?

grausof:
> can you add the reason to the PR / changeset description?

done!

apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
```diff
@@ -14,10 +14,8 @@ const issuer: Issuer = {
   id: newId(),
   subscriptionId: newId(),
   email: "info@enpacl-pec.it" as EmailString,
-  address: "Viale Del Caravaggio, 78 - 00147 Roma (RM)",
   description: "descrizione dell'ente" as NonEmptyString,
-  taxCode: "80119170589",
-  vatNumber: "80119170589",
+  isTrying: true,
```
gunzip, line 18: I'd use some other name since `isTrying` is unclear (it could be a production tentative)

seomthing like `isTestEvent` or `isSandboxEvent`...

grausof: However, this is a property of the issuer, not of the event!

grausof: probably `isTesting` ?

gunzip: so we're assuming that _every_ product belonging to the same issuer can be in testing o production as a whole, isn't it?

gunzip: for me `isTesting` is ok, what about it @lucacavallaro ?

gunzip (changes requested):

apps/io-func-sign-issuer/src/__test__/signature-request.spec.ts
lucacavallaro: I honestly don't like it very much 🙁 can't we use a field structured like `pricing plan` explained [here](https://pagopa.atlassian.net/wiki/spaces/SFEQS/pages/552108033/Fatturazione)?

to me the `test phase` it's just an another plan

grausof: I like this solution. We could think of a `pricePlan` field which can be of type `TEST` or `STANDARD` for now.

gunzip: I would agree if this flag is only about the testing plan, but according to what @grausof said, it also routes the signature action to the QTSP test endpoint.

lucacavallaro: @gunzip we can call it `environment` instead of `pricing plan`. A test environment produces signatures with `free plan` and redirect requests to QTSP test endpoint

gunzip: Yep, better!

grausof: Ok, I renamed the field to environment allowing two values: `TEST` or `DEFAULT`

gunzip (approved):

PR #70, opened 2023-01-30
feat: [SFEQS-1303]  Send billing events to data-lake via event-hubs

grausof:
This PR adds event management. In particular it adds the sending of billing events on an event-hub. Please refer to the Billing documentation on confluence.

Depends on https://github.com/pagopa/io-infra/pull/396

#### List of Changes
Added `event` type to package `io-sign`
Added `azure/infra/event-hubs` integration on `io-func-sign-issuer`
Added dispatch of the billing event in the `mark-request-signed` use-case
Added `EventHubsConnectionString` environment variable

#### Checklist:


packages/io-sign/src/event.ts
```diff
@@ -0,0 +1,47 @@
+import { IsoDateFromString } from "@pagopa/ts-commons/lib/dates";
```
gunzip, line 1: what about having this a a directory containing the single `<event>.ts` instead?

packages/io-sign/src/event.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Id, newId } from "./id";
+import {
+  SignatureRequestId,
+  SignatureRequestSigned,
+} from "./signature-request";
+
+const EventId = Id;
+
+// We currently only have a free testing plan and a paid standard plan
+export const PricingPlan = t.union([t.literal("FREE"), t.literal("DEFAULT")]);
```
gunzip, line 13: Consider the usage of `t.keyOf`

https://github.com/gcanti/io-ts/tree/1.x#union-of-string-literals

gunzip (approved):

apps/io-func-sign-issuer/src/app/use-cases/mark-request-signed.ts
```diff
@@ -24,5 +26,19 @@ export const makeMarkRequestAsSigned =
...
+      TE.chain(() =>
+        pipe(
+          request,
+          /*
+          The plan to use varies according to the environment used by the issuer.
+          If it is in a test environment the free plan should be used otherwise standard.
+          */
+          createBillingEvent(
+            request.issuerEnvironment === "TEST" ? "FREE" : "DEFAULT",
+            request.issuerId
```
lucacavallaro, line 39: imho this logic should be placed inside `createBillingEvent`

or maybe you can introduce a new function called `createBillingEventFromIssuer` that compose `createBillingEvent` with this logic

packages/io-sign/src/event.ts
lucacavallaro, line 13: use `t.keyof` instead of `t.union` https://github.com/gcanti/io-ts/blob/master/index.md#union-of-string-literals

lucacavallaro (changes requested):

packages/io-sign/src/event.ts
grausof: For all other entities it was done this way inside `io-sign`. If we change for event we should also change for other entities.

lucacavallaro: @gunzip Why do you recommend making a directory? Personally, I have no issues with dividing up our domain's directories if doing so makes the project simpler to navigate.

I believe that `event.ts` is now too short to be split up into several files; but, as this module grows, we may decide to reevaluate this in the future.

apps/io-func-sign-issuer/src/app/use-cases/mark-request-signed.ts
grausof: I created a new function `createBillingEventFromIssuer` but I had to add `getIssuerById` in `infra/azure/cosmos/issuer`

grausof: I moved this check inside `createBillingEvent`

lucacavallaro (approved):

PR #75, opened 2023-02-06
[SFEQS-1279]: Implementazione Azure identity per Github OIDC

manuelcoppotelli:
#### List of Changes

Defining an Azure Federated Identity binded to this repository

#### Motivation and Context

This repository can use Github Action to perform deploy on Azure Function so that enables continuous deployment

- tf validate
- tfsec
- tf plan prod

#### Checklist:

#### Other information

Depends on pagopa/io-infra#417


manuelcoppotelli:
@pagopa/infrastructure-admins could you please review the `.identity` directory and apply it?

lucacavallaro:
@manuelcoppotelli can you create a top-level directory where we can store the whole terraform code? something called `infra` or `tf`

I don't think that a `.`-prefixed directory it's the best place, usually this prefix is used for "dotfiles"/configuration files/repo metadata

manuelcoppotelli:
@lucacavallaro that will be the next step. Please notice that the terraform files inside the directory are not aimed at creating infrastructure resources relative to the domain, instead they are just "binding" this Github repo to an Azure subscription.

In a certain way, we are "configuring" this Github repo to communicate with Azure, therefore I think the dot prefix would make sense.

lucacavallaro:
> @lucacavallaro that will be the next step. Please notice that the terraform files inside the directory are not aimed at creating infrastructure resources relative to the domain, instead they are just "binding" this Github repo to an Azure subscription.
> 
> In a certain way, we are "configuring" this Github repo to communicate with Azure, therefore I think the dot prefix would make sense.

I'm not convinced :( in my opinion it could all be inside `tf/repo`, `infra/github`, `infra/repo` etc. I believe that `.identity` is not even easy to understand at first glance.

manuelcoppotelli:
> I'm not convinced :( in my opinion it could all be inside `tf/repo`, `infra/github`, `infra/repo` etc. I believe that `.identity` is not even easy to understand at first glance.

For technical reasons due to how terraform works, these tf files cannot be in the same directory as the other terraform files; they cannot be applied by an automation and they must be run separately that the other infra resources (they cannot be applied in a continuous deployment way).

Therefore my strong recommendation is to keep them segregated, from the infra as well

lucacavallaro (approved):

pasqualedevita (commented):
If we can wait, I'll review this PR after migration to azurerm v3 (https://github.com/pagopa/io-infra/pull/422) and container app runner creation.

I see possibile improvements to identity permissions because it doesn't need permissions on entire subscription but only on some functions or resource group.
Also I think that the CI Action (alias pull request) does not need access to Azure so we can remove this identity

pasqualedevita:
@lucacavallaro .identity is only for trust this repository with Azure with a new identity. We don't create new infrastructure resources.

.gitignore
```diff
@@ -28,3 +27,44 @@ __azurite_db_queue_extent__.json
...
+
+# .tfstate files
+*.tfstate
+*.tfstate.*
+
+# Crash log files
+crash.log
+
+# Ignore any .tfvars files that are generated automatically for each Terraform run. Most
+# .tfvars files are managed as part of configuration and so should be included in
```
pasqualedevita, line 41: for more portable terraform code, you can create a .gitignore into .identity dir, like https://github.com/pagopa/io-functions-services/tree/opex-demo/.identity

Also is recommended to add .terraform-version inside .identity

manuelcoppotelli: Agreed

gunzip (approved):

PR #76, opened 2023-02-08
chore: [SFEQS-1350] Add maxPollingInterval for queues

grausof:
This PR sets the `maxPollingInterval` for queues to 5 seconds and try to solve the problem of waiting more than one minute for the signature request status to change from `READY` to `WAIT_FOR_SIGNATURE`.

The [queuetrigger](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue-trigger) implements a random exponential back-off algorithm. As a result of this if there are no messages on the queue, the SDK will back off and start polling less frequently.

`MaxPollingInterval` allows us to configure this behavior. It for when a queue remains empty, the longest period of time to wait before checking for a message to.

### DIFF
--- a/.changeset/ten-hornets-dream.md
+++ b/.changeset/ten-hornets-dream.md
@@ -1,0 +1,6 @@
+---
+"io-func-sign-issuer": patch
+"io-func-sign-user": patch
+---
+
+Set maxPollingInterval for queues to 5 seconds
--- a/apps/io-func-sign-issuer/host.json
+++ b/apps/io-func-sign-issuer/host.json
@@ -5,7 +5,7 @@
       "routePrefix": "api/v1/sign"
     },
     "queues": {
-      "maxPollingInterval": "00:00:01.000"
+      "maxPollingInterval": "00:00:05.000"
     }
   }
 }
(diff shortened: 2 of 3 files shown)


gunzip (approved):
lgtm but I suspect that the polling interval defaults to 30s for a reason (ie. to reduce calls from the "serverless" function), maybe @manuelcoppotelli knows more.

manuelcoppotelli:
The `MaxPollingInterval` is set fro two reasons:

1. The `GetMessages` invocation has a cost, small but it is not free... thus reduce the bill at the end of the month. (this is also the reason why by default messages are retrieved in batch of 16 at a time)
2. Prevent the function from being flooded from messages; so to make the function "rest" between two consequent batches.

As a consequence, since we have very quite few messages and quite distant in time, we do not benefit from batch processing (because the message is probably "alone" in the queue), but instead we should wait for the polling interval.

Long story short, when we will have a more consistent volume of traffic, we could think of increasing this value back to its original default (i.e. 1 minute) without much pain.

gunzip:
This is a good point, 1 minute interval let us batch more messages and reduce invocations.
I guess we should track this into the backlog 🙏

PR #78, opened 2023-02-16
feat: [SFEQS-1404] Implement endpoints to use messages with attachments

grausof:
This PR implements the 2 endpoints needed to use messages with attachments.

In particular, the following endpoints have been implemented:
- `/messages/{signatureRequestId}` which allows you to obtain the list of documents (attachments) associated with a signature request.
- `/messages/{signatureRequestId}/{documentId}` Which allows you to obtain the content of the single document in the `application/pdf` format

#### Checklist:

### DIFF
--- a/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
+++ b/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
@@ -1,12 +1,12 @@
 import { pipe } from "fp-ts/lib/function";
 import * as TE from "fp-ts/lib/TaskEither";
-import * as A from "fp-ts/lib/Array";
 
 import { GetDocumentContent } from "@io-sign/io-sign/document-content";
 import { Document } from "@io-sign/io-sign/document";
 import { EntityNotFoundError } from "@io-sign/io-sign/error";
 
 import {
+  getDocument,
   SignatureRequest,
   signedNoMoreThan90DaysAgo,
 } from "../../signature-request";
@@ -20,8 +20,8 @@ export const makeGetSignedDocumentContent =
       TE.fromEither,
       TE.chain((signatureRequest) =>
         pipe(
-          signatureRequest.documents,
-          A.findFirst((document) => document.id === documentId),
+          signatureRequest,
+          getDocument(documentId),
           TE.fromOption(
             () =>
               new EntityNotFoundError(
--- a/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
+++ b/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
@@ -2,11 +2,12 @@ import { pipe } from "fp-ts/lib/function";
 import * as TE from "fp-ts/lib/TaskEither";
 
 import { GetDocumentContent } from "@io-sign/io-sign/document-content";
-import { Document } from "@io-sign/io-sign/document";
+import { Document, DocumentReady } from "@io-sign/io-sign/document";
 import { EntityNotFoundError } from "@io-sign/io-sign/error";
 
+import { getDocument } from "@io-sign/io-sign/signature-request";
+import { validate } from "@io-sign/io-sign/validation";
 import {
-  getDocument,
   SignatureRequest,
   signedNoMoreThan90DaysAgo,
 } from "../../signature-request";
@@ -27,6 +28,9 @@ export const makeGetSignedDocumentContent =
               new EntityNotFoundError(
                 "The specified documentID does not exists."
               )
+          ),
+          TE.chainEitherKW(
+            validate(DocumentReady, "The document must be in READY status.")
           ),
           TE.chain(getDocumentContent)
         )
--- a/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
+++ b/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
@@ -47,8 +47,7 @@ export const makeGetSignedDocumentContent =
       TE.chain((signatureRequest) =>
         pipe(
           signatureRequest.documents,
-          A.filter((document) => document.id === documentId),
-          A.head,
+          A.findFirst((document) => document.id === documentId),
           TE.fromOption(
             () =>
               new EntityNotFoundError(
--- a/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
+++ b/apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
@@ -1,41 +1,15 @@
 import { pipe } from "fp-ts/lib/function";
-import * as E from "fp-ts/lib/Either";
 import * as TE from "fp-ts/lib/TaskEither";
 import * as A from "fp-ts/lib/Array";
-
-import { differenceInDays } from "date-fns";
 
 import { GetDocumentContent } from "@io-sign/io-sign/document-content";
 import { Document } from "@io-sign/io-sign/document";
 import { EntityNotFoundError } from "@io-sign/io-sign/error";
-import { SignatureRequestSigned } from "@io-sign/io-sign/signature-request";
-import { validate } from "@io-sign/io-sign/validation";
 
-import { SignatureRequest } from "../../signature-request";
-
-export const signedNoMoreThan90DaysAgo = (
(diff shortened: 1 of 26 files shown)


apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
```diff
@@ -0,0 +1,60 @@
...
+          difference < 90
+            ? E.right(signatureRequest)
+            : E.left(
+                new EntityNotFoundError(
+                  "More than 90 days have passed since signing."
+                )
+              )
+      )
+    )
+  );
```
lucacavallaro, line 38: this logic should be placed at the `domain` level

apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
```diff
@@ -0,0 +1,60 @@
...
+  (getDocumentContent: GetDocumentContent) =>
+  (signatureRequest: SignatureRequest, documentId: Document["id"]) =>
+    pipe(
+      signatureRequest,
+      signedNoMoreThan90DaysAgo,
+      TE.fromEither,
+      TE.chain((signatureRequest) =>
+        pipe(
+          signatureRequest.documents,
+          A.findFirst((document) => document.id === documentId),
```
lucacavallaro, line 50: here you should use the `getDocument` function from the domain

apps/io-func-sign-user/src/infra/azure/functions/get-third-party-message-attachments-content.ts
```diff
@@ -0,0 +1,98 @@
...
+    error,
+    (buffer) => ({
+      // body must be of type string, but buffer.toString appends some extra characters which corrupt the final file even with byte-encoding.
+      body: buffer as unknown as string,
+      statusCode: 200,
+      headers: {
+        "Content-Type": "application/pdf",
+        "Content-Length": Buffer.byteLength(buffer).toString(),
+      },
+    })
```
lucacavallaro, line 91: the response should be constructed via the `handler-kit` utility functions

apps/io-func-sign-user/src/infra/http/decoder/fiscal-code.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import * as E from "fp-ts/lib/Either";
+
+import { HttpBadRequestError } from "@io-sign/io-sign/infra/http/errors";
+import { FiscalCode } from "@pagopa/ts-commons/lib/strings";
+
+export const requireFiscalCode = (req: HttpRequest) =>
+  pipe(
+    req,
+    header("fiscal_code"),
```
lucacavallaro, line 15: usually custom header names are prefixed (like `x-io-`) and written in `kebab-case` :'( 

if this choice does not depend on our system, maybe we can write a comment that explains the relation with external services

apps/io-func-sign-user/src/infra/http/decoder/signature-request.ts
```diff
@@ -42,3 +45,38 @@ export const makeRequireSignatureRequest = (
...
+              new EntityNotFoundError("The specified signer does not exists.")
+          )
+        ),
+        TE.map((signer) => signer.id),
+        TE.chain(getSignatureRequest(signatureRequestId))
+      )
+    ),
+    RTE.chainW(
+      RTE.fromOption(
+        // eslint-disable-next-line sonarjs/no-identical-functions
```
lucacavallaro, line 75: extract the function instead of disable `eslint`

apps/io-func-sign-user/src/infra/http/decoder/signature-request.ts
```diff
@@ -42,3 +45,38 @@ export const makeRequireSignatureRequest = (
...
     )
   );
+
+export const makeRequireSignatureRequestByFiscalCode = (
+  getSignatureRequest: GetSignatureRequest,
+  getSignerByFiscalCode: GetSignerByFiscalCode
+): RTE.ReaderTaskEither<HttpRequest, Error, SignatureRequest> =>
+  pipe(
+    sequenceS(RE.Apply)({
+      fiscalCode: flow(requireFiscalCode),
```
lucacavallaro, line 55:
```suggestion
      fiscalCode: requireFiscalCode,
```

lucacavallaro (changes requested):

apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
grausof: `getDocument` exist only in` io-func-sign-issuer`. I create it in `io-func-sign-user` as well

apps/io-func-sign-user/src/infra/azure/functions/get-third-party-message-attachments-content.ts
grausof: Ok I add a dependency to the PR!

apps/io-func-sign-user/src/infra/http/decoder/fiscal-code.ts
grausof: Unfortunately this choice does not depend on our system as specified [here](https://github.com/pagopa/io-backend/blob/c7257c946b99830fe8c052d4cc0d4dc78b000d51/openapi/consumed/api-third-party.yaml#L94). I add a comment!

apps/io-func-sign-user/src/infra/http/decoder/signature-request.ts
grausof: 🫢

apps/io-func-sign-user/src/infra/azure/functions/get-third-party-message-attachments-content.ts
grausof: I checked the `handler-kit `code better and it is general purpose so it can already support raw type responses as well. I try to move this logic into `@io-sign` like `success` -> `jsonResponse`

apps/io-func-sign-user/src/app/use-cases/get-signed-document-content.ts
lucacavallaro: Move it to '@io-sign/sign'

apps/io-func-sign-user/src/signature-request.ts
```diff
@@ -180,3 +185,27 @@ export const markAsRejected = (reason: string) =>
...
+    signatureRequest,
+    validate(
+      SignatureRequestSigned,
+      "The signature request must be in SIGNED status."
+    ),
+    E.chain((signatureRequest) =>
+      pipe(
+        differenceInDays(new Date(), signatureRequest.signedAt),
+        (difference) =>
+          difference < 90
```
gunzip, line 202: What about making this parametric (taken from a configuration object?)

gunzip (approved):

apps/io-func-sign-user/src/signature-request.ts
grausof: How about if we use an `envvar` to set the number of days for a signing request to be valid?

gunzip: yes, env var is ok, but you can only access it at top level or from the configuration object (not directly inside this method)

grausof: ok then I make sure that this value is read from a configuration, but instead of an `envvar` I insert it hard-coded in the [main](https://github.com/pagopa/io-sign/blob/main/apps/io-func-sign-user/src/app/main.ts) as done for other parameters (such as the names of the collections)

lucacavallaro: imho this should not be an `env variable`, it's a simple "domain invariant" for the signature request entity

If we think it should change over time it is better to track it on the database

grausof: @lucacavallaro what do you think about putting this invariant in `main.ts`?

lucacavallaro: ok let's leave it as it is now without changes, if in the future we need to change it we will understand why, how and any implications

lucacavallaro (approved):

PR #79, opened 2023-02-17
basic host.json config for functions

manuelcoppotelli:
#### Checklist:


apps/io-func-sign-issuer/host.json
```diff
@@ -1,9 +1,31 @@
...
+    "applicationInsights": {
+      "samplingSettings": {
+        "isEnabled": true,
+        "maxTelemetryItemsPerSecond": 5,
+        "includedTypes": "PageView;Trace;Dependency;Request",
+        "excludedTypes": "Exception;Event;CustomEvent"
+      }
+    }
+  },
+  "extensionBundle": {
```
gunzip, line 17: we'd either use a budle OR a custom configuration of packages like -> https://github.com/pagopa/io-sign/blob/main/apps/io-func-sign-issuer/extensions.csproj

apps/io-func-sign-user/host.json
```diff
@@ -1,9 +1,31 @@
...
+    "applicationInsights": {
+      "samplingSettings": {
+        "isEnabled": true,
+        "maxTelemetryItemsPerSecond": 5,
+        "includedTypes": "PageView;Trace;Dependency;Request",
+        "excludedTypes": "Exception;Event;CustomEvent"
+      }
+    }
+  },
+  "extensionBundle": {
```
gunzip, line 17: ditto

gunzip: see https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=v4%2Cmacos%2Ccsharp%2Cportal%2Cbash#install-extensions

grausof:
Settled in https://github.com/pagopa/io-sign/pull/80

apps/io-func-sign-issuer/host.json
manuelcoppotelli: It might be useful to consider that `extensions.csproj` file is recommended in case of manual installation and portal-only development[^1].

We should also consider that, when using `extensions.csproj`, other than requiring the extension to be explicitly installed, the `extensions.csproj` is ignored[^2] in case of `extensionBundle` property is defined in `host.json`.

Therefore I would go with the `host.json` solution.


[^1]: <https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#explicitly-install-extensions>
[^2]: <https://github.com/MicrosoftDocs/azure-docs/issues/52767#issuecomment-675661019>

manuelcoppotelli:
Thank you @grausof for integrating the `host.json` files within the proper PR.

Has anyone alto planned to integrate the remaining files (e.g. `.devcontainer`, `repsonse.ts`)?

PR #81, opened 2023-02-20
chore: [SFEQS-1419] Update signed message to support third party content

grausof:
This PR adds support for `third party content` and modifies the `message` of successful signature.

#### List of Changes
- Added `third_party_data` as an optional parameter in `@io-sign/io-sign/notification`
- Updated message of successful signature

#### Checklist:


packages/io-sign/src/infra/io-services/message.ts
```diff
@@ -36,6 +37,11 @@ export const makeSubmitMessageForUser =
               message: {
                 ...message,
                 fiscal_code: fiscalCode,
+                /* feature_level_type field is used to identify the institutions that have subscribed to premium messages.
+                 * In our case we have not adhered to any agreement therefore the field remains STANDARD but
+                 * in any case we are enabled to use the attachments feature.
```
gunzip, line 42: how?

lucacavallaro (changes requested):

packages/io-sign/src/infra/io-services/message.ts
gunzip: I'd add this to the comment

packages/io-sign/src/notification.ts
```diff
@@ -11,12 +12,23 @@ export const Notification = t.type({
...
     subject: t.string,
     markdown: t.string,
   }),
+  t.partial({
+    third_party_data: t.type({
+      id: NonEmptyString,
+      has_attachments: t.boolean,
+    }),
+  }),
+]);
```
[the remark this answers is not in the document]
grausof: Yes, I was thinking about doing a refactor like this. And then maybe with an encoder on the infra side. I proceed and ask you for a review!

packages/io-sign/src/infra/io-services/message.ts
```diff
@@ -15,6 +19,25 @@ import { ActionNotAllowedError, TooManyRequestsError } from "../../error";
...
+export const NotificationMessageToApiModel: Enc.Encoder<
+  NewMessage,
+  NotificationMessage
+> = {
+  encode: (message) => ({
+    content: {
+      subject: message.subject,
+      markdown: message.markdown,
+      third_party_data:
+        "signatureRequestId" in message
```
gunzip, line 31: What about having two different encoders instead? 
(`NotificationContent` vs `NotificationContentWithAttachments`)

grausof: Done! it's a good idea!

packages/io-sign/src/notification.ts
```diff
@@ -11,13 +12,27 @@ export const Notification = t.type({
...
+  t.type({
+    signatureRequestId: Id,
+  }),
+]);
+
+export type NotificationContentWithAttachments = t.TypeOf<
+  typeof NotificationContentWithAttachments
+>;
+
+export type NotificationMessage =
```
gunzip, line 33: if you use a io-ts union instead you can call `NotificationMessage.encode` without first checking for the `signatureId`.

grausof: It's not clear to me. How do I define more than one encoder for the same type? NotificationMessage is part of the io-sign domain

gunzip: you're right sorry :-)

lucacavallaro (approved):

PR #82, opened 2023-02-21
fix: [SFEQS-1431] Add filename extension to ThirdPartyMessageApiModel

grausof:
Add `.pdf` extension to filename in third party message attachments.


apps/io-func-sign-user/src/infra/http/encoders/signature-request.ts
```diff
@@ -88,7 +90,9 @@ export const SignatureRequestToThirdPartyMessage: E.Encoder<
       A.map((document) => ({
         id: document.id,
         content_type: "application/pdf" as NonEmptyString,
-        name: document.metadata.title,
+        name: pipe(document.metadata.title, S.endsWith(".pdf"))
```
gunzip, line 93: I'd extract this into its own method

gunzip (approved):

PR #85, opened 2023-02-27
Demo issuer partial generated cli

filippo-tenaglia:
This PR will contain the source code of the demo created using the openapi generator tool with the flag typescript.
The folder contains many files autogenerated as described on the readme.md found on the parent folder and few files that should integrated the cli's logic.
These files are:
main.Cli.ts and signer-cli.ts.
At this moment if you tried the code the terminal return this error after the calling to the first API implemented:
ReferenceError: XMLHttpRequest is not defined

Can @gunzip  help me to understand where is the error?

Here its are the steps to reproduce it:
1) npm install
2) tsc
3) npm run start-cli
4) select "signer"
5) insert any character tho execute the call.

Tho exit from the cli you can choose any other APIs.


gunzip:
@filippo-tenaglia the generated code is utilizing the [whatwg-fetch fetch](https://www.npmjs.com/package/whatwg-fetch) package; it's designed for web browsers, which is the cause of the problem. To make the generator build for NodeJS instead, you probably need to feed it some sort of command line flag. Let me review the other PR with the ADR to make sure we're aligned on the code generating tool before moving on.

filippo-tenaglia:
@gunzip  I resolved the problem updating node to the latest LTS version (18.14) now the API response with a 422 error code but response!

filippo-tenaglia:
@gunzip  I solved the issue of the empty body. Now I'm trying to refactor the code because on SignerCli.ts there are too many promises!

filippo-tenaglia:
@gunzip  in this last commit I added all references already tested on ts-cli probject so now you can call all APIs correctly. There is, probably, an error during the creation of a new signature request because the class SignatureField autogenerated seems have to need both the coords and the unique name! so, If you want to try to create a new signature request, select the unique name option during the creation of the clauses.

filippo-tenaglia:
@lucacavallaro  and @grausof  I modified the .gitignore file to add on the repository the folder models autogenerated by openapi generator

docs/demo-issuer/openapi-gen-ts/src/generated/models/SignatureField.ts
```diff
@@ -0,0 +1,44 @@
+/**
```
filippo-tenaglia, line 1: @gunzip  @lucacavallaro  @grausof  using this class seems that I must pass both unique_name and attrs with some values, can you see if it is correct or I didn't see something?
The consequence of this behaviour can create problems on the backend right?

gunzip:
closed in favor of https://github.com/pagopa/io-sign/pull/91

PR #87, opened 2023-03-03
feat!: [SFEQS-1463] Self-care integration

grausof:
This PR adds self-care integration.
Depends on https://github.com/pagopa/io-infra/pull/472

#### List of Changes

- Added `CreateIssuer` function triggered by selfcare event-hub
- Added `SelfCareEventHubConnectionString` to envvar
- Added `internalInstitutionId` parameter to `Issuer`
- Added connection to `self-care` eventhub for issuer creation after institution onboarding
- Fixed sending `internalInstitutionId` to `data-lake` for billing
- Added `internalInstitutionId` to `SignatureRequest` (introduces a https://github.com/pagopa/io-sign/labels/breaking-change on existing signature requests)

#### Motivation and Context
With this PR, an issuer is created automatically as soon as the selfcare onoboarding ends.

Tested locally with cloud resources and UAT selfcare event-hub

#### Checklist:


apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
```diff
@@ -70,6 +75,34 @@ class IssuerModel extends CosmosdbModel<
...
+        () =>
+          pipe(
+            this.getQueryIterator({
+              parameters: [
+                {
+                  name: "@internalInstitutionIdKey",
+                  value: internalInstitutionId,
+                },
+              ],
+              query: `SELECT * FROM m WHERE m.${internalInstitutionIdKey} = @internalInstitutionIdKey`,
```
gunzip, line 92:
```suggestion
              query: `SELECT TOP 1 * FROM m WHERE m.${internalInstitutionIdKey} = @internalInstitutionIdKey`,
```

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
```diff
@@ -70,6 +75,34 @@ class IssuerModel extends CosmosdbModel<
...
+    internalInstitutionId: string
+  ): TE.TaskEither<CosmosErrors, O.Option<RetrievedIssuer>> {
+    return pipe(
+      TE.tryCatch(
+        () =>
+          pipe(
+            this.getQueryIterator({
+              parameters: [
+                {
+                  name: "@internalInstitutionIdKey",
```
gunzip, line 88: is this the partitionkey?

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuer.ts
```diff
@@ -0,0 +1,52 @@
...
+  GenericContracts,
+  IoSignContract,
+} from "../../self-care/contract";
+import { ioSignContractToIssuer } from "../../self-care/encoder";
+import {
+  makeCheckIssuerWithSameInternalInstitutionId,
+  makeInsertIssuer,
+} from "../cosmos/issuer";
+
+const makeValidateUploadHandler = (db: CosmosDatabase) => {
```
gunzip, line 22: why "upload"? maybe a comment can help here

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuer.ts
```diff
@@ -0,0 +1,52 @@
...
+  const insertIssuer = makeInsertIssuer(db);
+
+  const createIssuerFromContract = flow(
+    contractActive,
+    E.chainW(validate(IoSignContract, "This is not an `io-sign` contract")),
+    E.map(ioSignContractToIssuer.encode),
+    TE.fromEither,
+    TE.chain(checkIssuerWithSameInternalInstitutionId),
+    TE.chain(insertIssuer),
+    // This is a fire-and-forget operation
```
gunzip, line 39: are you sure you don't want to retry in case the insert operation fails?

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
gunzip: I don't remember but maybe there's some cosmosdb method (on this) that you can use to get an item from its id.

apps/io-func-sign-issuer/src/infra/self-care/encoder.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as E from "io-ts/lib/Encoder";
+import * as S from "fp-ts/lib/string";
+import { pipe } from "fp-ts/lib/function";
+import { newId } from "@io-sign/io-sign/id";
+import { IoSignContract } from "./contract";
+
+export const ioSignContractToIssuer: E.Encoder<Issuer, IoSignContract> = {
+  encode: ({ internalIstitutionID, institution }) => ({
+    id: newId(),
+    // See [IO-SIGN-RFC-SUB-ID]
```
gunzip, line 12: can you elaborate this comment a little bit more? (what's happening here?)

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
grausof: No, this is the parameter to look up!

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuer.ts
grausof: it's a typo!

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuer.ts
grausof: The insert operation could fail only in case the DB is not responding. We should be able to figure out the type of error to perform a retry. Maybe I'll add a TODO

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
grausof: Here we don't have to get an item from the id but from a parameter so i think the query is needed.

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
gunzip: yu must include the partition key in every query

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuer.ts
gunzip: yes, every IO operation that could fail (ie. db not responding) must be retried, that's why you must make your handler idempotent.

gunzip (changes requested):

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
grausof: In this case we do not know the `partitionKey` as the selfCare issuer id is contained only in the `internalInstitutionId` field. Therefore, a full table scan is required to understand if an entity already exists.

gunzip: this is a problem. when you omit the partition key every partition will be queried, so if you have 10 partitions, this means 10 different http calls to cosmosdb. you probably want to use just one partition if the table is small (I think this is doable someway with cosmos)

grausof: A simple idea could be to use the same selfCare ID (`internalInstitutionId`) also for the issuer ID. In this way we don't have insertion problems as the id will be unique. In my opinion, having an `issuerId` and an `internalInstitutionId` makes no sense as there cannot be an institution with the same issuerId and a different internalInstitutionId.

grausof: [here](https://github.com/pagopa/io-sign/blob/607dbc7b564352b03610458d5dbe580d95c7c38d/apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts#L73)  too there is the same problem

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
```diff
@@ -69,33 +73,30 @@ class IssuerModel extends CosmosdbModel<
...
-      TE.map(RA.flatten),
-      TE.chainW(
-        flow(E.sequenceArray, E.mapLeft(CosmosDecodingError), TE.fromEither)
+export const makeGetIssuerByVatNumber =
+  (db: cosmos.Database): GetIssuerByVatNumber =>
+  (vatNumber: Issuer["vatNumber"]) =>
+    pipe(
+      new IssuerByVatNumberModel(db),
+      (model) => model.find([vatNumber]),
+      TE.chain(
```
gunzip, line 84: maybe `chainOptionK` can help?

lucacavallaro (approved):

PR #89, opened 2023-03-06
feat(issuer): [SFEQS-1351] Require at least one mandatory signature

lucacavallaro:
#### List of Changes

1. Created a new codec for `SignatureFields` that checks the "require at least one mandatory signature" constraint

The change has been unit tested

#### Checklist:

### DIFF
--- a/packages/io-sign/src/__test__/document.spec.ts
+++ b/packages/io-sign/src/__test__/document.spec.ts
@@ -3,16 +3,23 @@ import { describe, it, expect } from "@jest/globals";
 import { identity, pipe } from "fp-ts/lib/function";
 import * as E from "fp-ts/lib/Either";
 import {
+  NonEmptyString,
+  WithinRangeString,
+} from "@pagopa/ts-commons/lib/strings";
+import {
   DocumentMetadata,
   newDocument,
   startValidation,
   markAsReady,
   markAsRejected,
+  SignatureFields,
+  SignatureField,
 } from "../document";
+import { newId } from "../id";
 
 const metadata: DocumentMetadata = {
   title: "my test document",
-  signatureFields: [],
+  signatureFields: [] as unknown as DocumentMetadata["signatureFields"],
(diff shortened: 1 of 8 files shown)


packages/io-sign/src/document.ts
```diff
@@ -90,9 +90,23 @@ export const PdfDocumentMetadata = t.type({
...
+export interface SignatureFieldsBrand {
+  readonly SignatureFields: unique symbol;
+}
+
+export const SignatureFields = t.brand(
+  t.array(SignatureField),
+  (
+    signatureFields
+  ): signatureFields is t.Branded<SignatureField[], SignatureFieldsBrand> =>
+    signatureFields.length === 0 ||
```
grausof, line 102: It seems to me that if the list of signature fields is empty, however, we can proceed with the signature on the entire file!

packages/io-sign/src/__test__/document.spec.ts
```diff
@@ -94,3 +101,55 @@ describe("Document", () => {
...
+        attributes: {
+          uniqueName: title as NonEmptyString,
+        },
+        clause: {
+          title: title as WithinRangeString<5, 80>,
+          type,
+        },
+      };
+    };
+    const testTable: Array<{
```
gunzip, line 125: you probably can do this using `it.each`

gunzip (approved):

packages/io-sign/src/document.ts
lucacavallaro: yes, this check ensures that!

the signature field list must be empty or with at least one mandatory signature

grausof: OK!

grausof (approved):
LGTM

packages/io-sign/src/__test__/document.spec.ts
lucacavallaro: I tried to write with `it.each`, I confirm that it makes add types much more difficult

PR #94, opened 2023-03-13
feat(user): [SFEQS-1478,SFEQS-1479]  Lollipop integration to `createSignature` endpoint 🍭 

grausof:
This PR adds lollipop implementation on the `createSignature` endpoint inside `io-func-sign-user`

#### List of Changes
- Added headers required by lollipop to `createSignature`
- Added lollipop infrastructure for API access
- Added SPID assertion retrieval via lollipop API
- Implemented forwarding of parameters to QTSP
- Upgrated `@pagopa/openapi-codegen-ts` and `@pagopats-common`

Tested locally

#### Checklist:


apps/io-func-sign-user/src/app/use-cases/create-signature.ts
```diff
@@ -175,46 +166,25 @@ export const makeCreateSignature =
...
-          ),
-        })
-      ),
+      TE.map(({ documentsToSign, fiscalCode }) => ({
+        fiscalCode,
+        publicKey: signatureValidationParams.publicKeyBase64,
+        spidAssertion: Buffer.from(
+          signatureValidationParams.samlAssertion,
+          "utf-8"
+        ).toString("base64") as NonEmptyString,
```
lucacavallaro, line 176: I think it's better to do a runtime validation (`validate(NonEmptyString)`) against the result of `Buffer.from`

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,59 @@
...
+import { LollipopSignatureInput } from "../http/models/LollipopSignatureInput";
+
+const getSignRegexFromSignatureId = (signatureId: string) =>
+  new RegExp(`${signatureId}:[A-Za-z0-9+/=]*:?`);
+
+/* Given a LollipopSignature es: `sig1=:SIGNATURE_1:,sig2=:SIGNATURE_2:,sig3=:....` and a signatureInput like:
+ * `sig1=("content-type").....,sig2=("CUSTOM_HEADER_NAME_1");created=...,sig3=("CUSTOM_HEADER_NAME_2");created=...`
+ * return the signature (es: SIGNATURE_2) associated with the single header name (es: CUSTOM_HEADER_NAME_1)
+ */
+export const getSignatureFromHeaderName = (
```
gunzip, line 21: this method needs unit tests

gunzip (changes requested):

apps/io-func-sign-user/src/infra/lollipop/signature.ts
grausof: Are [here](https://github.com/pagopa/io-sign/pull/94/files#diff-4692cabafede767ad3fed30fac381c926511b8273864c7b0f7aed36f14147f7eR1)

apps/io-func-sign-user/src/app/use-cases/create-signature.ts
grausof: Ok I added validation on both `signatureInput` and `samlAssertion`

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,59 @@
...
+  signatureInput: LollipopSignatureInput,
+  sigantures: LollipopSignature,
+  headerName: string
+) =>
+  pipe(
+    signatureInput,
+    S.split(","),
+    RA.filterMapWithIndex((index: number, signatureInput: string) =>
```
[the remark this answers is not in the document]
grausof: done!

apps/io-func-sign-user/src/infra/http/decoder/lollipop.ts
```diff
@@ -0,0 +1,103 @@
...
+          ),
+          signChallenge: pipe(
+            request,
+            requireParameterFromHeader(
+              "x-pagopa-lollipop-custom-sign-challenge",
+              NonEmptyString
+            )
+          ),
+        }),
+        E.map((createSignatureLollipopParams) => ({
```
lucacavallaro, line 97: can you annotate here the static type that you expect after this map?

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,64 @@
...
+ * `sig1=("content-type").....,sig2=("CUSTOM_HEADER_NAME_1");created=...,sig3=("CUSTOM_HEADER_NAME_2");created=...`
+ * return the signature (es: SIGNATURE_2) associated with the header name (es: CUSTOM_HEADER_NAME_1)
+ */
+export const getSignatureFromHeaderName = (
+  signatureInput: LollipopSignatureInput,
+  signatures: LollipopSignature,
+  headerName: string
+) =>
+  pipe(
+    /* signatureInput field contains the parameters of all the signatures (sig1, sig2, ...) separated by comma
```
gunzip, line 27: What I meant was: describe each passage (in its own comment). ie. 

// Given a signature input: sig1=("FooBar"),sig2=("DeadBeef")
// and the header name: FooBar
// return the signature prefix with index: sig1
...

// Given a signature string: sig1=SomeValue;sig2=SomeOtherValue
// and the signature prefix with index: sig1
// extract the value part: SomeValue
...

// Given the signature value: SomeValue
// .. I don't know what's happening with S.split(/[::]/)
...

gunzip: I suspect that the corret regex is 

`new RegExp(`${signaturePrefix}:([^:])+`);`

gunzip (changes requested):

apps/io-func-sign-user/src/infra/lollipop/signature.ts
gunzip: `new RegExp("sig1=:([^:]+)").exec("sig1=:asdsad:")[1]`

grausof: Ok, I probably wasn't clear. I've tried reducing the code by adding a group to the regex. I hope with the added comments the flow is clearer now.

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import * as RA from "fp-ts/lib/ReadonlyArray";
+
+import { validate } from "@io-sign/io-sign/validation";
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+import { LollipopSignature } from "../http/models/LollipopSignature";
+import { LollipopSignatureInput } from "../http/models/LollipopSignatureInput";
+
+const getSignRegexFromSignaturePrefix = (signaturePrefix: string) =>
+  new RegExp(`${signaturePrefix}:([^:]*):?`);
```
gunzip, line 15:
```suggestion
  new RegExp(`${signaturePrefix}:([^:]+)`);
```

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,64 @@
...
+    // and the signature prefix with index: sig1=
+    // extract a regex match: sig1=:SomeValue: and a regex group: SomeValue
+    O.chain(
+      flow(
+        getSignRegexFromSignaturePrefix,
+        (re) => re.exec(signatures),
+        O.fromNullable
+      )
+    ),
+    // Given the regex result extracts the first group: SomeValue
```
gunzip, line 48: you don't need chain + filterWithIndex, just use `A.lookup(1)`

apps/io-func-sign-user/src/infra/lollipop/signature.ts
grausof: https://regex101.com/r/dYceap/1

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,64 @@
...
+    O.chain(
+      flow(
+        RA.filterWithIndex((group) => group === 1),
+        RA.head
+      )
+    ),
```
gunzip:
```suggestion
        RA.lookup(1),
```

gunzip (changes requested):

apps/io-func-sign-user/src/infra/lollipop/signature.ts
grausof: Thanks! 
Regex ninja 🥷

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,59 @@
...
+    // and the signature prefix with index: sig1=
+    // extract a regex match: [sig1=:SomeValue:, SomeValue]
+    O.chain(
+      flow(
+        getSignRegexFromSignaturePrefix,
+        (re) => re.exec(signatures),
+        O.fromNullable
+      )
+    ),
+    // Given the regex result extracts the first group: SomeValue
```
gunzip, line 48: describe the supposed result with an example, not the implementation which is auto-evident

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,59 @@
...
+    S.split(","),
+    RA.filterMapWithIndex((index: number, signatureInput: string) =>
+      pipe(signatureInput, S.startsWith(`sig${index + 1}=("${headerName}")`))
+        ? O.some(`sig${index + 1}=`)
+        : O.none
+    ),
+    RA.head,
+    // Given a signatures string: sig1=:SomeValue:,sig2=:SomeOtherValue:
+    // and the signature prefix with index: sig1=
+    // extract a regex match: [sig1=:SomeValue:, SomeValue]
```
gunzip, line 40:
```suggestion
    // extract the value part: SomeValue
```

gunzip (changes requested):

apps/io-func-sign-user/src/infra/lollipop/signature.ts
grausof: Ok, but I had to move` RA.lookup(1)` after `O.fromNullable`

apps/io-func-sign-user/src/infra/lollipop/signature.ts
lucacavallaro: couldn't tests be used as examples?

grausof: @lucacavallaro I think the aim is to explain step by step what happens and not the final result!

gunzip: it helps (even: code reviews) to co-locate comments, tests are about the whole implementation, here we describe each step. I'd to reverse engineer the code (and regex) to understand what's happening and make some contribution.

lucacavallaro: the whole implementation could be split into several (single responsibility) functions, which can be unit-tested individually if it helps to have better clarity

grausof: I think in this particular case maybe just add a unit-test just on the regex!

grausof: I split the `getSignatureFromHeaderName` function into 2 other functions (`getSignaturePrefixFromHeaderName`, `getSignatureFromPrefix`) and for each one I added a unit-test. I also defined the `LollipopSignaturePrefix` type for additional validation.

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/* Given a signatures string: sig1=:SomeValue:,sig2=:SomeOtherValue:
+ * and the signature prefix with index: sig1=
+ * extract the value part: SomeValue
+ */
+export const getSignatureFromPrefix =
+  (signatures: LollipopSignature) =>
+  (signaturePrefix: LollipopSignaturePrefix) =>
+    pipe(
+      getSignRegexFromSignaturePrefix(signaturePrefix),
```
gunzip, line 52: just inline the regex here

apps/io-func-sign-user/src/infra/lollipop/signature.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const getSignRegexFromSignaturePrefix = (
+  signaturePrefix: LollipopSignaturePrefix
+) => new RegExp(`${signaturePrefix}:([^:]+)`);
+
+/* Given a signature input: sig1=("FooBar"),sig2=("DeadBeef")
+ * and the header name: FooBar
+ * return the signature prefix with index: sig1=
+ */
+export const getSignaturePrefixFromHeaderName =
```
gunzip, line 26: you can do this with a regex as well, something like

```
const getSignaturePrefix = (signatureInput: string) => (headerName: string) =>
  pipe(
    new RegExp('(sig[0-9]+=)\\("' + headerName + '"\\)').exec(signatureInput),
    O.fromNullable,
    O.chain(RA.lookup(1))
  );

```

PR #95, opened 2023-03-14
fix: Lollipop public key encoding

grausof:
This PR fix encoding of publicKey!

### DIFF
--- a/.changeset/seven-vans-exercise.md
+++ b/.changeset/seven-vans-exercise.md
@@ -1,0 +1,5 @@
+---
+"io-func-sign-user": patch
+---
+
+Fixed public key encoding


apps/io-func-sign-user/src/app/use-cases/create-signature.ts
```diff
@@ -195,6 +195,24 @@ export const makeCreateSignature =
           }))
         )
       ),
+      TE.chainEitherKW((createSignaturePayload) =>
+        pipe(
+          Buffer.from(createSignaturePayload.publicKey, "base64").toString(
+            "utf-8"
+          ),
+          JSON.parse,
```
gunzip, line 203: use fp-ts json parse (this may throw)

PR #96, opened 2023-03-15
feat(issuer): [SFEQS-1505] Add CreateIssuerByVatNumberView function

lucacavallaro:
#### List of Changes

1. Add a new function, triggered by cosmos db, called `CreateIssuerByVatNumberView`
2. Add `vatNumber` property on `Issuer` entity

#### Motivation and Context

This function creates a "lookup" collection on CosmosDB that allows us to get issuers by their vat number instead of `(issuer.id, subscription.id)`

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/CreateIssuerByVatNumberView/function.json
+++ b/apps/io-func-sign-issuer/CreateIssuerByVatNumberView/function.json
@@ -1,0 +1,26 @@
+{
+  "bindings": [
+    {
+      "type": "cosmosDBTrigger",
+      "name": "documents",
+      "direction": "in",
+      "leaseCollectionName": "issuers-leases",
+      "connectionStringSetting": "CosmosDbConnectionString",
+      "databaseName": "%CosmosDbDatabaseName%",
+      "collectionName": "issuers",
+      "createLeaseCollectionIfNotExists": true,
+      "startFromBeginning": true
+    },
+    {
+      "type": "cosmosDB",
+      "direction": "out",
+      "connectionStringSetting": "CosmosDbConnectionString",
+      "databaseName": "%CosmosDbDatabaseName%",
+      "collectionName": "issuers-by-vat-number",
+      "createIfNotExists": false,
+      "name": "$return"
+    }
+  ],
+  "scriptFile": "../dist/main.js",
+  "entryPoint": "CreateIssuerByVatNumberView"
+}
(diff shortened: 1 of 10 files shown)


grausof (approved):
Good job!

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuers-by-vat-number-view.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as TE from "fp-ts/TaskEither";
+
+import { validate } from "@io-sign/io-sign/validation";
+
+import * as t from "io-ts";
+import { RetrievedIssuer } from "../cosmos/issuer";
+
+export const run: AzureFunction = (_, documents: unknown) =>
+  pipe(
+    documents,
```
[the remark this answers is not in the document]
lucacavallaro: renamed to `items` ✅ 

(when we'll update `handler-kit`, this raw handler will be replaced by a new `itemHandler`)

apps/io-func-sign-issuer/CreateIssuerByVatNumberView/function.json
```diff
@@ -0,0 +1,26 @@
+{
+  "bindings": [
+    {
+      "type": "cosmosDBTrigger",
+      "name": "items",
+      "direction": "in",
+      "leaseCollectionName": "issuers-leases",
```
gunzip, line 7: consider that the lease collection is usually one per cosmosdb instance (shared by all functions) since when the function writes the lease it can use the prefix setting `leaseCollectionPrefix` (so you don't have to create a new container for each lease):

```
"leaseCollectionName": "leases",
"leaseCollectionPrefix": "issuers",

```

apps/io-func-sign-issuer/src/infra/azure/functions/create-issuers-by-vat-number-view.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as E from "fp-ts/Either";
+import * as A from "fp-ts/Array";
+import * as TE from "fp-ts/TaskEither";
+
+import { validate } from "@io-sign/io-sign/validation";
+
+import * as t from "io-ts";
+import { RetrievedIssuer } from "../cosmos/issuer";
+
+export const run: AzureFunction = (_, items: unknown) =>
```
gunzip, line 13: can you add a comment on what this function does?

lucacavallaro: i think it's already very explicit, the filename is litterally `create issuers by vat number view`

I'll add a comment that explains the `WHY` (or the context) behind this view

PR #98, opened 2023-03-16
ci: [SFEQS-949] Migrazione pipeline di deploy a GitHub actions

manuelcoppotelli:
#### List of Changes

- configured `changesets` tools to handle private packages
- release function can build and create github releases with artifacts
- added github action to deploy on azure via self-hosted runners

#### Motivation and Context

Automate the build and the release processes

#### Checklist:

### DIFF
--- a/.github/workflows/azure-deploy.yaml
+++ b/.github/workflows/azure-deploy.yaml
@@ -4,6 +4,12 @@ on:
   release:
     types: [published]
 
+env:
+  ENV_SHORT: p
+  FUNCTION_SLOT: staging
+  PREFIX: io
+  RESOURCE_GROUP: io-p-sign-backend-rg
+
 jobs:
   create_runner:
     name: Create Runner
@@ -33,13 +39,17 @@ jobs:
       - name: Download artifact
         run: "curl --output artifact.zip --location ${{ github.event.release.assets[0].browser_download_url }}"
 
-      - name: Identity workload
+      - name: Identify workload
         id: workload
         shell: bash
         run: |
+          # Parse the release artifact name to extract type, domain and name
           set -euo pipefail
+
           echo type=$(echo "${{ github.event.release.assets[0].name }}" | sed -En "s/^io-([a-z0-9]+)-([a-z0-9]+)-([a-z0-9]+)-.+$/\1/p" | tr -d '\n') >> "$GITHUB_OUTPUT"
+
           echo domain=$(echo "${{ github.event.release.assets[0].name }}" | sed -En "s/^io-([a-z0-9]+)-([a-z0-9]+)-([a-z0-9]+)-.+$/\2/p" | tr -d '\n') >> "$GITHUB_OUTPUT"
+
           echo name=$(echo "${{ github.event.release.assets[0].name }}" | sed -En "s/^io-([a-z0-9]+)-([a-z0-9]+)-([a-z0-9]+)-.+$/\3/p" | tr -d '\n') >> "$GITHUB_OUTPUT"
 
       - name: Login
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -11,20 +11,96 @@ jobs:
...
+            yarn.lock
+            package.json
+            apps/**/package.json
+            packages/**/package.json
+
+      - name: Install dependencies
+        run: yarn
+
+      - name: Generate TypeScript models from OpenAPI
+        run: yarn workspace ${{ matrix.package.name }} run generate:api-models
```
gunzip, line 92: imho this is a task that should run automatically when launching the `build` task (e.g., a `prebuild` npm task) this is doable changing `package.json`. Otherwise you (the developer) must know in advance the right sequence to call before build cc @lucacavallaro

.github/workflows/azure-deploy.yaml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          client-id: ${{ env.AZURE_CLIENT_ID }}
+          tenant-id: ${{ env.AZURE_TENANT_ID }}
+          subscription-id: ${{ env.AZURE_SUBSCRIPTION_ID }}
+
+      - name: Function
+        if: steps.workload.outputs.type == 'func'
+        shell: bash
+        run: |
+          az webapp deploy -g io-p-sign-backend-rg -n io-p-${{ steps.workload.outputs.domain }}-${{ steps.workload.outputs.name }}-func -s staging --type zip --src-path artifact.zip
```
gunzip, line 58: can we isolate variables (ie. `io-p-sign-backend-rg`) in some section at the beginning? this would help to know the movimng parts in case someone want to replicate the action somewhere else.

.github/workflows/azure-deploy.yaml
```diff
@@ -0,0 +1,77 @@
...
+    needs: create_runner
+    steps:
+      - name: Download artifact
+        run: "curl --output artifact.zip --location ${{ github.event.release.assets[0].browser_download_url }}"
+
+      - name: Identity workload
+        id: workload
+        shell: bash
+        run: |
+          set -euo pipefail
```
gunzip, line 40: what's happening here? :)

gunzip (commented):
looks promising!

.github/workflows/azure-deploy.yaml
manuelcoppotelli: Making sure the whole bash step will fail even in one of the intermediate piped command fails

.github/workflows/azure-deploy.yaml
manuelcoppotelli: Sure we can

.github/workflows/azure-deploy.yaml
gunzip: I mean, in the whole script

.github/workflows/release.yaml
lucacavallaro: I don't agree, in my opinion tasks should do only one thing and shouldn't have "side effect".

Also I think it's inefficient to generate the models on every build.

In the future we'll add to the repository the generated files :)

gunzip: Code generation is not a side effect, but a fundamental step of the build (at least until generated files are committed).

.github/workflows/azure-deploy.yaml
```diff
@@ -0,0 +1,77 @@
...
+        id: create_github_runner
+        # from https://github.com/pagopa/github-self-hosted-runner-azure-create-action/commits/main
+        uses: pagopa/github-self-hosted-runner-azure-create-action@63534a04613b420ce6474ccbb52ac7884129ff6f
+        with:
+          client_id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant_id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription_id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          container_app_environment_name: ${{ secrets.AZURE_CONTAINER_APP_ENVIRONMENT_NAME }}
+          resource_group_name: ${{ secrets.AZURE_RESOURCE_GROUP_NAME }}
+          pat_token: ${{ secrets.BOT_TOKEN }}
```
grausof, line 25: sure we can not use `${{ secrets.GITHUB_TOKEN }} `?

manuelcoppotelli: Yes, tested and cannot work due to the restriction of the API which is supposed to call.

.github/workflows/release.yaml
manuelcoppotelli: @lucacavallaro please let me know in case the building commands should change

grausof (approved):
LGTM

PR #100, opened 2023-03-16
feat: [SFEQS-1517] Add slack notification for new issuers

grausof:
This PR adds `slack` integration to notify new issuers onboarding via `self-care`.

Depends on: https://github.com/pagopa/io-infra/pull/489
#### List of Changes
1. Adds client for `slack`
2. Add `postSlackMessage` function

#### Motivation and Context
It is necessary to report the onboarding of new entities in order to proceed with the sending of the api-keys.


apps/io-func-sign-issuer/src/infra/slack/config.ts
```diff
@@ -23,3 +23,7 @@ export const getSlackConfigFromEnvironment: RE.ReaderEither<
   ),
   apiToken: readFromEnvironment("SlackApiToken"),
 });
+
+export const slackChannelMap = {
+  si_firmaconio_tech: "C03G0KJBU7N",
```
lucacavallaro, line 28: to me this id should be read from an `env` var 

i'll remove the channel "map" and I'll replace it with just the `slackChannelId`

apps/io-func-sign-issuer/src/infra/slack/config.ts
```diff
@@ -3,14 +3,14 @@ import * as t from "io-ts";
 import * as RE from "fp-ts/lib/ReaderEither";
 import { sequenceS } from "fp-ts/lib/Apply";
 import { pipe } from "fp-ts/lib/function";
-import { readFromEnvironment } from "../env";
+import { readFromEnvironment } from "@io-sign/io-sign/infra/env";
 
 export const SlackConfig = t.type({
   apiBasePath: t.string,
   apiToken: t.string,
```
lucacavallaro, line 10: the `apiToken` is really needed? thought we could use [https://api.slack.com/messaging/webhooks](https://api.slack.com/messaging/webhooks) for a lighter integration

apps/io-func-sign-issuer/src/infra/slack/client.ts
lucacavallaro, line 1: maybe we can generalize some of the function and structures in these clients (`self care` client and `slack client` looks too similar)

lucacavallaro (changes requested):

apps/io-func-sign-issuer/src/infra/slack/config.ts
grausof: if we use webhooks it is no longer necessary to have a channelId!

packages/io-sign/src/infra/client-utils.ts
```diff
@@ -0,0 +1,28 @@
+import * as t from "io-ts";
+
+import * as E from "fp-ts/lib/Either";
+import { flow, pipe } from "fp-ts/lib/function";
+import * as TE from "fp-ts/lib/TaskEither";
+import { readableReport } from "@pagopa/ts-commons/lib/reporters";
+
+export const isSuccessful = (r: Response): boolean =>
```
lucacavallaro, line 8: it's almost perfect, but I ask you for one last (_optional, for now_) effort: instead of validating (`isSuccessful`) it would be better to parse the response to make it a `SuccessfulResponse` (using a `type guard`) 

the reasons for this pattern can be found explained here https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

gunzip (approved):

PR #106, opened 2023-03-27
feat (user): [SFEQS-1501] Add "GetSignatureRequests" function

lucacavallaro:
#### List of Changes

1. Add `GetSignatureRequests` function that returns the list of the signature requests created by a signer
2. Refactor code to import `@pagopa/handler-kit@1.0.0` instead of "legacy"-unstable version
3. Change log settings (the default is now "Error" for functions)

#### Motivation and Context

This endpoint is needed for the "support flow" on the user side

The new endpoint and the new decoder has been unit tested

#### Checklist:


apps/io-func-sign-user/src/infra/azure/cosmos/signature-request.ts
```diff
@@ -71,3 +86,42 @@ export const makeUpsertSignatureRequest =
       (model) => model.upsert(signatureRequest),
       TE.mapLeft(toCosmosDatabaseError)
     );
+
+export class CosmosDbSignatureRequestRepository
+  implements SignatureRequestRepository
+{
+  #model: SignatureRequestModel;
```
grausof, line 93: can you explain this syntax with a comment?

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
```diff
@@ -29,7 +29,7 @@ import {
 } from "@io-sign/io-sign/infra/azure/storage/blob";
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import { makeCreateFilledDocumentUrl } from "../../../app/use-cases/create-filled-document";
-import { requireSigner } from "../../http/decoder/signer";
+import { requireSigner } from "../../http/decoder/signer.old";
```
grausof, line 32: are we introducing technical debt?

apps/io-func-sign-user/src/infra/http/encoders/signature-request.ts
```diff
@@ -86,6 +88,46 @@ export const SignatureRequestToApiModel: E.Encoder<
   },
 };
 
+export const toSignatureRequestStatusEnum = (
+  status: SignatureRequest["status"]
+): SignatureRequestStatusEnum => {
```
grausof, line 93: Here I would specify that it is the state of `SignatureRequestListView` and not of `SignatureRequest`.If not necessary we can also eliminate the export.

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
lucacavallaro: we can't refactor all handlers right now! It will take some time, so for now we have to duplicate decoders

when all handlers will be refactored to the latest version of handler-kit we can remove the `*.old` files

hope that we can migrate before the launch :)

apps/io-func-sign-user/src/infra/http/encoders/signature-request.ts
lucacavallaro: nope, this state is also the state of `SignatureRequest` (see the change in `openapi.yaml`)

lucacavallaro: (I'll remove the export!)

PR #109, opened 2023-03-27
feat (user): [SFEQS-1315] Adds support to the QTSP test environment

grausof:
This PR adds support to the QTSP test environment therefore depending on the issuer environment it is chosen which QTSP environment to use (`TEST`/`DEFAULT`)

#### List of Changes
- Added issuer-environment parameter in `getSignatureRequest`
- Added `x-iosign-issuer-environment` header to `getQtspClauses`
- Added QTSP test environment

#### Checklist:


apps/io-func-sign-user/openapi.yaml
```diff
@@ -753,9 +763,12 @@ components:
             description:
               type: string
               minLength: 1
+            environment:
+              $ref: "#/components/schemas/IssuerEnvironment"
           required:
             - email
             - description
+            - environment
```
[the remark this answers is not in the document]
grausof: this is a response with an added parameter. Why do you think it's a breaking change?

apps/io-func-sign-user/src/app/use-cases/create-signature.ts
```diff
@@ -167,84 +167,95 @@ export const makeCreateSignature =
...
+          TE.chainEitherKW((createSignaturePayload) =>
+            pipe(
+              Buffer.from(createSignaturePayload.publicKey, "base64").toString(
+                "utf-8"
+              ),
+              JSON.parse,
+              (jsonKey) =>
+                Buffer.from(JSON.stringify(jsonKey), "utf-8").toString(
+                  "base64"
+                ),
```
lucacavallaro, line 211: can you extract this conversion to base64 code in a dedicated function?

lucacavallaro (changes requested):

apps/io-func-sign-user/src/app/use-cases/create-signature.ts
grausof: why do you think it is necessary to write an ad-hoc function for the conversion? Perhaps it would be better to extrapolate the stringify part of the json

grausof: I added a utility function in the @io-sign package!

packages/io-sign/src/utility.ts
```diff
@@ -0,0 +1,22 @@
+import * as E from "fp-ts/lib/Either";
+import { pipe } from "fp-ts/lib/function";
+
+export const stringToBase64Encode = (arrayBuffer: string) =>
```
lucacavallaro, line 4: (additional tip) since you are using `Either` you can also decode the input as a string (so you can take an `unknown` as input ad avoid `undefined as unknown as string`)

lucacavallaro (approved):

packages/io-sign/src/utility.ts
grausof: done!

PR #111, opened 2023-03-28
Demo issuer partial generated cli v3

filippo-tenaglia:
This pull request should contain the tool to test, and not only, the Issuer's API.
@gunzip  The three methods seems work correctly but only one return the right result, can you help me to find the problem?
To test you can execute node dist/index.js and you will see, as response, two "undefined" and one json object with an ID, the two undefined should return:
the new signature request just created;
and the dossier passed by ID.


gunzip:
I've pushed a fix for DocumentMetadata (it has changed). I need an API key to test this.

packages/io-sign-issuer-sdk/src/signature-request.ts
filippo-tenaglia, line 1: @gunzip at line 30 the compiler return: "The right-hand side of an 'instanceof' expression must be of type 'any' or of a type assignable to the 'Function' interface type." I'm quite sure that there is some stupid error but I can't find it!
@lucacavallaro  about the MD sent yesterday, I was looking for "validation class" to do this check ... what did you mean with your answer about parsing? Is it correct in this context?

gunzip: I see no errors and the project build successfully. You probably need to build all the other packages as well

packages/io-sign-issuer-sdk/src/middleware.ts
```diff
@@ -0,0 +1,15 @@
...
+  FetchParams,
+} from "@io-sign/io-sign-api-client";
+
+export class APIMiddleware implements Middleware {
+  pre(context: ResponseContext): Promise<FetchParams | void> {
+    return Promise.resolve({ url: context.url, init: context.init });
+  }
+
+  post(context: ResponseContext): Promise<Response | void> {
+    return Promise.resolve(context.response);
```
filippo-tenaglia, line 13: @gunzip  as I told you, some API calls returns undefined some not, I think that the problem is here because the Response object contains one, or more, promises as child so the printed result could be not complete, can you help me to find where is the definition of this class? (I can't find it on API client project)

PR #112, opened 2023-04-03
test(issuer): [SFEQS-1554] Add unit test to dossier endpoints

lucacavallaro:
#### List of Changes

1. Migrate `CreateDossier` and `GetDossier` to `handler-kit@1`
2. Add unit test

unit test + manual e2e test in development environment

#### Checklist:


apps/io-func-sign-issuer/src/issuer.ts
```diff
@@ -1,20 +1,53 @@
...
+  getBySubscriptionId: (
+    subscriptionId: Issuer["subscriptionId"]
+  ) => TE.TaskEither<Error, O.Option<Issuer>>;
+};
+
+type IssuerEnvironment = {
+  issuerRepository: IssuerRepository;
+};
+
+const getIssuerByField =
```
gunzip, line 23: I'd written two different methods instead.

apps/io-func-sign-issuer/src/infra/http/decoders/issuer.ts
```diff
@@ -1,30 +1,61 @@
...
+    ),
+    E.chainW(
+      H.parse(
+        Issuer.props.subscriptionId,
+        "The content of x-subscription-id is not a valid id"
+      )
+    )
+  );
+
+export const requireIssuer = flow(
```
gunzip, line 56: 1) @silvicir this is a good example of a Reader(TaskEither) usage. 

If you look at the legacy implementation above, you may spot that `makeRequireIssuer` takes an already "configured" mehotd (`getIssuerBySubscriptionId`) as parameter. 
This means that every time you have to use `makeRequireIssuer` you have to provide a working instance of `getIssuerBySubscriptionId` (which needs an `IssuerRepository` as a dependency, and so on).

When you want to use `requireIssuer`, which returns a Reader instead, (same implementation of `makeRequireIssuer`) you can defer the "plumbing" of dependencies to some other place (usually: the main method). Look at the method `CreateDossierHandler` below.

apps/io-func-sign-issuer/src/infra/http/handlers/create-dossier.ts
```diff
@@ -0,0 +1,53 @@
...
+        ),
+      })
+    ),
+    fromEither
+  );
+
+export const CreateDossierHandler = H.of((req: H.HttpRequest) =>
+  pipe(
+    sequenceS(ApplyPar)({
+      issuer: requireIssuer(req),
```
gunzip, line 43: 2) Here we're using `requireIssuer` even if it calls `getIssuerBySubscriptionId` which is not yet configured with the `IssuerRepository`.

apps/io-func-sign-issuer/src/app/main.ts
```diff
@@ -159,3 +158,16 @@ export const CreateIssuer = makeCreateIssuerFunction(
...
+const issuerRepository = new CosmosDbIssuerRepository(database);
+const dossierRepository = new CosmosDbDossierRepository(database);
+
+export const GetDossier = GetDossierFunction({
+  issuerRepository,
+  dossierRepository,
+});
+
+export const CreateDossier = CreateDossierFunction({
+  issuerRepository,
```
gunzip, line 171: 3) Only here we finally pass the `IssuerRepository` (inside main file) @silvicir

apps/io-func-sign-issuer/src/issuer.ts
lucacavallaro: yep, you right: it's a bit ugly right now! I'll split it in two (more linear) functions

apps/io-func-sign-issuer/src/infra/http/decoders/issuer.ts
```diff
@@ -1,30 +1,61 @@
...
 import { HttpUnauthorizedError } from "@io-sign/io-sign/infra/http/errors";
-import { GetIssuerBySubscriptionId } from "../../../issuer";
+import {
+  getIssuerBySubscriptionId,
+  GetIssuerBySubscriptionId,
+} from "../../../issuer";
 
-const requireSubscriptionId = (req: HttpRequest) =>
+// ------- DECODERS FOR HANDLER-KIT LEGACY ----------- //
+// this block can be removed when the entire app has been upgrated to handler-kit@1
```
grausof, line 22: add a TODO ref

apps/io-func-sign-issuer/src/infra/azure/cosmos/issuer.ts
```diff
@@ -75,6 +76,44 @@ class IssuerModel extends CosmosdbModel<
...
+      TE.mapLeft(toCosmosDatabaseError)
+    );
+  }
+}
+
+// LEGACY FUNCTIONS
+// This block can be removed when the entire app has been ported to handler-kit@1
 export const makeGetIssuerByVatNumber =
   (db: cosmos.Database): GetIssuerByVatNumber =>
   (vatNumber: Issuer["vatNumber"]) =>
```
grausof, line 119: ditto. Add a TODO ref.

apps/io-func-sign-issuer/src/infra/azure/cosmos/dossier.ts
```diff
@@ -30,6 +35,27 @@ class DossierModel extends CosmosdbModel<
...
+  getById(id: Dossier["id"], issuerId: Dossier["issuerId"]) {
+    return pipe(
+      this.#model.find([id, issuerId]),
+      TE.mapLeft(toCosmosDatabaseError)
+    );
+  }
+}
+
+// LEGACY FUNCTIONS
+// This block can be removed when the entire app has been ported to handler-kit@1
```
grausof, line 58: ditto

apps/io-func-sign-issuer/src/dossier.ts
```diff
@@ -39,8 +44,42 @@ export const newDossier = (
...
+      repo.getById(id, issuerId),
+      TE.chain(
+        TE.fromOption(
+          () => new EntityNotFoundError("The specified dossier was not found")
+        )
+      )
+    );
+
+// LEGACY TYPES
+// This block can be removed when the entire app has been ported to handler-kit@1
```
grausof, line 80: ditto

grausof (approved):
LGTM. I would leave this PR open after the merge as a reference for the integration of the new handler-kit

PR #113, opened 2023-04-03
fix: [SFEQS-1580] Remove filledDocument if exist before creating it

grausof:
When createFilledDocument is called, if a previous filledDocument already exists, it is deleted.

#### List of Changes
- Added `deleteBlobIfExist` in io-sign package
- Added deletion of previous `filledDocument` in `createFilledDocument`

#### Motivation and Context
The presence of a previous document caused a bug when switching spid sessions where the attributes were different.

Locally

#### Checklist:

### DIFF
--- a/apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
+++ b/apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
@@ -26,6 +26,7 @@ import {
   withPermissions,
   withExpireInMinutes,
   getBlobClient,
+  deleteBlobIfExist,
 } from "@io-sign/io-sign/infra/azure/storage/blob";
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
 import { makeCreateFilledDocumentUrl } from "../../../app/use-cases/create-filled-document";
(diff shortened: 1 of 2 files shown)


lucacavallaro (approved):

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
```diff
@@ -54,6 +55,14 @@ const makeCreateFilledDocumentHandler = (
     pipe(
       filledDocumentBlobName,
       getBlobClient,
+      RTE.chainFirstTaskEitherK((blobClient) =>
```
gunzip, line 58: we probably want to pass this as a  `makeCreateFilledDocumentUrl` dep instead of calling it here directly inside `GetFilledDocumentUrl`

gunzip (changes requested):

apps/io-func-sign-user/src/infra/azure/functions/create-filled-document.ts
lucacavallaro: as we migrate to the new version of `@pagopa/handler-kit`, all these `makeX` functions will be removed

I don't know how much worth investing in this refactor now

gunzip: In which way they will be removed?

lucacavallaro: `makeX = (deps) => (payload) => TE`

will become 

`X = (payload) => (deps) => TE`

in order to be composable with other `RTEs`

PR #114, opened 2023-04-04
feat: [SFEQS-1583] Add log to `createSignature`

grausof:
Add log with `@pagopa/logger` to `createSignature` endpoint without refactor to new `handler-kit`

#### Motivation and Context
The refactor to the new handler-kit requires changing many files that we can avoid for now. This PR creates an adapter to use the new logger also on a legacy version of handler-kit

Tested locally

#### Checklist:


packages/io-sign/src/infra/console-logger.ts
```diff
@@ -0,0 +1,22 @@
+import * as L from "@pagopa/logger";
+
+// Derive a concrete implementation L.Logger using azure.Context.log
+export const ConsoleLogger: L.Logger = {
```
gunzip, line 4: isn't this suppsed to be taken from function context?

grausof: CC @lucacavallaro

grausof: I proposed [this change](https://github.com/pagopa/io-sign/pull/114/commits/36d6aeb58a0ba5b208b489187be0919ea902c8e0) to use context.log

.changeset/chilly-buses-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"io-func-sign-user": minor
+---
+
+Added consoleLog to createSignature
```
gunzip, line 5: I suggest to avoid putting implementation details in chagesets, but explain the reason of the change
```suggestion
Added logging to the create-signature http handler
```

.changeset/kind-pumpkins-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@io-sign/io-sign": patch
+---
+
+Added ConsoleLog
```
gunzip, line 5: ditto (these changelog should be understandable by anyone)

.changeset/metal-coats-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"io-func-sign-support": patch
+---
+
+Upgrated packages
```
gunzip, line 5:
```suggestion
Upgraded packages
```

(which packages?)

packages/io-sign/src/infra/console-logger.ts
gunzip: ping

grausof: @lucacavallaro says we are introducing another type of handler this way

.changeset/kind-pumpkins-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@io-sign/io-sign": patch
+---
+
+Added ConsoleLog utility function to log through console
```
gunzip, line 5:
```suggestion
Added utility function to log through console
```

gunzip (approved):

PR #115, opened 2023-04-05
feat!: [SFEQS-1570] Add department to issuer and INTERNAL pricing plan

grausof:
Adds the field `department` to `issuer` useful for billing with datalake.
⚠ Introduce breaking change on `signatureRequest` model

#### List of Changes
- Add `department` to `issuer`
- Add `INTERNAL` pricing plan

#### Motivation and Context
INTERNAL pricingPlan is introduced to manage the events of using internal real signatures but which do not need to be billed

The department field is a field that will be used to discriminate signatures by department in the invoice issued. (I'll give you an example, let's suppose that the University of Rome signs a contract for signature with IO, at this point a generic issuer is created for the university. But the university wants to distribute the product firma con io to different departments such as engineering or letters, therefore a new issuer must be created that has the same billing data, therefore with the same internalInstitutionId but with this department parameter which allows you to indicate which department that issuer belongs to).

Tested locally

#### Checklist:


lucacavallaro:
I don't understand why we are introducing these fields. We should be relying on environment and fields for billing, could you give us more context?

grausof:
> I don't understand why we are introducing these fields. We should be relying on environment and fields for billing, could you give us more context?

So let me try to summarize:
 - the environment field is used to indicate whether the issuer is in a test phase or not, this is used to discriminate whether fake or real signatures should be used
 - the isInternal field is used to distinguish whether an entity is internal to PagoPa or external.  Although PagoPa's internal entities can consume both test signatures and real signatures, they must not be counted for the purposes of the invoice but in any case the consumed signatures (those for a fee) must be counted by datalake in order to generate a quarterly report on the signatures consumed by us and that the QTSP will invoice
 - The department field, on the other hand, is a field that will be used to discriminate signatures by department in the invoice issued.  (I'll give you an example, let's suppose that the University of Rome signs a contract for signature with IO, at this point a generic issuer is created for the university. But the university wants to distribute the product firma con io to different departments such as engineering or letters, therefore a new issuer must be created that has the same billing data, therefore with the same internalInstitutionId but with this department parameter which allows you to indicate which department that issuer belongs to)
 I hope I was clear,
 Thank you

lucacavallaro:
> * the isInternal field is used to distinguish whether an entity is internal to PagoPa or external.  Although PagoPa's internal entities can consume both test signatures and real signatures, they must not be counted for the purposes of the invoice but in any case the consumed signatures (those for a fee) must be counted by datalake in order to generate a quarterly report on the signatures consumed by us and that the QTSP will invoice

ok, but why we can't introduce a new plan for internal consumer? we should provide to datalake an unique way to count signatures

grausof:
> > * the isInternal field is used to distinguish whether an entity is internal to PagoPa or external.  Although PagoPa's internal entities can consume both test signatures and real signatures, they must not be counted for the purposes of the invoice but in any case the consumed signatures (those for a fee) must be counted by datalake in order to generate a quarterly report on the signatures consumed by us and that the QTSP will invoice
> 
> 
> 
> ok, but why we can't introduce a new plan for internal consumer? we should provide to datalake an unique way to count signatures

We had proposed it but I'll try again.

grausof:
> > > * the isInternal field is used to distinguish whether an entity is internal to PagoPa or external.  Although PagoPa's internal entities can consume both test signatures and real signatures, they must not be counted for the purposes of the invoice but in any case the consumed signatures (those for a fee) must be counted by datalake in order to generate a quarterly report on the signatures consumed by us and that the QTSP will invoice
> > 
> > 
> > ok, but why we can't introduce a new plan for internal consumer? we should provide to datalake an unique way to count signatures
> 
> We had proposed it but I'll try again.

I eliminated the `isInternal` field and added a new environment: `INTERNAL`

gunzip (approved):

PR #117, opened 2023-04-07
chore: [SFEQS-1596] Add console log support to host.json

grausof:
Adds support to the `console.log ` in the `host.json` file

#### Motivation and Context
The use of console.log is different from the context.log therefore logging via console must also be enabled.

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/host.json
+++ b/apps/io-func-sign-issuer/host.json
@@ -2,8 +2,8 @@
   "version": "2.0",
   "logging": {
     "logLevel": {
-      "Function": "Error",
-      "Function.Console": "Information"
+      "Function": "Debug",
+      "Function.Console": "Debug"
     },
     "applicationInsights": {
       "samplingSettings": {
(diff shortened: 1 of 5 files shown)


apps/io-func-sign-user/host.json
```diff
@@ -2,8 +2,8 @@
   "version": "2.0",
   "logging": {
     "logLevel": {
-      "default": "None",
-      "Function": "Error"
+      "Function": "Debug",
+      "Function.Console": "Debug"
```
manuelcoppotelli, line 6: Let's be conservative and tune the log in a successive moment

manuelcoppotelli (approved):

apps/io-func-sign-issuer/host.json
```diff
@@ -2,8 +2,8 @@
   "version": "2.0",
   "logging": {
     "logLevel": {
-      "default": "None",
-      "Function": "Error"
+      "Function": "Debug",
```
gunzip, line 5: this will result in a very verbose output with latency issues, are we sure is this what we want?

PR #118, opened 2023-04-07
feat(opex): SFEQS-1357 Implement operational excellence dashboards

manuelcoppotelli:
#### Checklist:


.opex/api_user/env/prod/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+location: West Europe
+timespan: 10m
+data_source: /subscriptions/ec285037-c673-4f58-b594-d7c480da4e8b/resourceGroups/io-p-rg-external/providers/Microsoft.Network/applicationGateways/io-p-appgateway
+resource_type: app-gateway
+action_groups:
+  - /subscriptions/ec285037-c673-4f58-b594-d7c480da4e8b/resourceGroups/io-p-rg-common/providers/microsoft.insights/actionGroups/SlackPagoPA
+  - /subscriptions/ec285037-c673-4f58-b594-d7c480da4e8b/resourceGroups/io-p-rg-common/providers/microsoft.insights/actionGroups/EmailPagoPA
+overrides:
+  hosts:
+    - app-backend.io.italia.it
```
gunzip, line 12: The specification file for the backend API is [this one](https://github.com/pagopa/io-backend/blob/master/api_io_sign.yaml) which is a little bit different from the one that is exposed by the function (ie. it does not handles some headers). Anyway, since the endpoints are the same, probably it's ok to configure the dashboard as is (just double check this).

apps/io-func-sign-issuer/openapi.yaml
```diff
@@ -542,7 +542,6 @@ components:
           $ref: "#/components/schemas/Timestamp"
       required:
         - id
```
gunzip, line 544: why this change here?

apps/io-func-sign-user/openapi.yaml
```diff
@@ -4,6 +4,8 @@ info:
   version: 0.0.1
 security:
   - FunctionsKey: []
+servers:
+  - url: https://app-backend.io.italia.it/api/v1/sign
```
gunzip, line 8: I'm not sure this is right (see previous comment).

gunzip: (this probably belongs to the io-backend dashboard instead)

apps/io-func-sign-issuer/openapi.yaml
lucacavallaro: @gunzip because the old value was wrong! neither swagger UI nor our generator reported this, but Manuel noticed it while he was integrating the OPEX dashboard (if you open the file you can see that the `status` property does not exists on `Document`)

apps/io-func-sign-user/openapi.yaml
grausof: yes, the address is wrong. The openapi specification of that endpoint is different and contains fewer parameters.

grausof (approved):

.opex/api_user/env/prod/config.yaml
manuelcoppotelli: Ok, it is correct to monitor form the client prospective.

apps/io-func-sign-user/openapi.yaml
manuelcoppotelli: Thanks. This will be corrected one referencing the correct openapi file

.opex/api_user/env/prod/config.yaml
manuelcoppotelli: The openapi file indicated needed some fixes as well. We shoudl wait for https://github.com/pagopa/io-backend/pull/1005 to be merged

PR #119, opened 2023-04-11
feat: [SFEQS-1352] Add analytics events

grausof:
Adds analytics event implementation for datalake

#### List of Changes
- Added signature request creation event

#### Motivation and Context
Refs:
https://pagopa.atlassian.net/wiki/spaces/SFEQS/pages/552108033/Fatturazione
https://pagopa.atlassian.net/wiki/spaces/SFEQS/pages/554238606/Strategy+Metrics

Locally with cloud resources

#### Checklist:


apps/io-func-sign-issuer/src/infra/azure/functions/create-signature-request.ts
```diff
@@ -113,7 +122,16 @@ const makeCreateSignatureRequestHandler = (db: CosmosDatabase) => {
 
   return createHandler(
     decodeHttpRequest,
-    createSignatureRequest,
+    flow(
+      createSignatureRequest,
+      TE.chain((signatureRequest) =>
+        pipe(
+          signatureRequest,
+          createAndSendAnalyticsEvent(EventName.SIGNATURE_CREATED),
```
gunzip, line 130: what if the `createSignatureRequest` succeeds and `createAndSendAnalyticsEvent` fails? we probably want one between:

A. retry the whole operation (ensure that the implementation is idempotent)
B. continue anyway (`createAndSendAnalyticsEvent` becomes _fire and forget_)

grausof: In this case the retry can be done directly by the caller of the signature request creation function. However I think that for this kind of events it is better to make them all fire and forget (if it is lost nothing happens). 

I try to add a fold directly on the `createAndSendAnalyticsEvent`

grausof: I made this method asynchronous!

packages/io-sign/src/infra/azure/event-hubs/event.ts
```diff
@@ -0,0 +1,80 @@
...
+            L.error("Unable to send analytics event", {
+              eventName,
+              signatureRequest,
+            })({
+              logger: ConsoleLogger,
+            })
+          )
+        )
+      )
+    )();
```
gunzip, line 78: ensure that in case of rejection / exceptions this get caught, otherwsie the whole process will be killed (it's usually enough to add a `catch` clause)

packages/io-sign/src/infra/azure/event-hubs/event.ts
```diff
@@ -0,0 +1,80 @@
...
+    signatureRequest:
+      | SignatureRequestDraft
+      | SignatureRequestSigned
+      | SignatureRequestReady
+      | SignatureRequestToBeSigned
+      | SignatureRequestWaitForQtsp
+      | SignatureRequestRejected
+  ): TE.TaskEither<Error, typeof signatureRequest> => {
+    // This is a fire and forget async operation
+    void pipe(
```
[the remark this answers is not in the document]
grausof: reverted to A 🤯

gunzip (commented):
lgtm but lacks unit tests

grausof:
> lgtm but lacks unit tests
@gunzip 
without the integration of the new handler-kit it is not possible to add unit-tests to this type of implementation!

gunzip:
we should at least test that the method returned by `makeCreateAndSendAnalyticsEvent` does not return a left value

grausof:
> we should at least test that the method returned by `makeCreateAndSendAnalyticsEvent` does not return a left value

Done, check if it is as you expected them!

gunzip (approved):

PR #129, opened 2023-04-17
feat: [SFEQS-1352] Add analytics event

grausof:
Adds sending of analytics events to `datalake` eventhub (PDV).

#### List of Changes
Added the following analytics events:
-  Document uploaded 
- Document rejected
- Signature ready
- Signature send
- Notification rejected
- Signature signed
- Signature rejected
- QTSP errors

TEsted locally

#### Checklist:


apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
```diff
@@ -78,7 +80,27 @@ export const makeSendNotification =
...
+              firstTaskEither,
+              TE.altW(() =>
+                pipe(
+                  signatureRequest,
+                  createAndSendAnalyticsEvent(EventName.NOTIFICATION_REJECTED),
+                  TE.chain(() => firstTaskEither)
+                )
+              )
+            )
+        ),
```
lucacavallaro, line 103: it's not very linear, try to rewrite it using only one `createAndSendAnalyticsEvent` at the end, the only thing that changes is the event name

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
```diff
@@ -105,15 +110,63 @@ export const makeValidateSignature =
...
+                pipe(
+                  firstTaskEither,
+                  TE.alt(() =>
+                    pipe(
+                      signatureRequest,
+                      createAndSendAnalyticsEvent(EventName.QTSP_API_ERROR),
+                      () => firstTaskEither
+                    )
+                  )
+                )
```
lucacavallaro, line 124: ditto (maybe here you can use `TE.chainFirst`)

lucacavallaro (changes requested):

apps/io-func-sign-user/src/app/use-cases/validate-signature.ts
grausof: Anfortunately for `Alt` there is no `TE.altFirst` lik e`TE.chainFirst` so this is the only way to do it.

apps/io-func-sign-issuer/src/app/use-cases/send-notification.ts
grausof: it is not possible to do this as the first time is a chain while the second call is inside an Alt.

lucacavallaro (approved):

PR #131, opened 2023-04-17
refactor(issuer): [SFEQS-1555] Refactor Validate Upload (+ logs)

lucacavallaro:
#### List of Changes

1. Ported `ValidatedUpload` to `handler-kit 1.x` (TE -> RTE)
2. Refactored to avoid nested `pipe` and eccessive branching
4. Add logs!

⚠️ there are other code smells that I will fix with subsequent PRs. I will also add the unit tests on ValidateDocument (which are partial at the moment).


grausof (approved):
It all seems correct to me. I trust unit tests

PR #133, opened 2023-04-19
feat: [SFEQS-1620] Add document validation analytics event 

grausof:
Add analytics events for accepted or rejected document. It also implements the `createAndSendAnalyticsEvent` as an RTE to support the new handler-kit.

#### List of Changes
- Added `createAndSendAnalyticsEvent` as RTE
- Added DOCUMENT_UPLOADED and DOCUMENT_REJECTED analytics events
- Converted the old `makeCreateAndSendAnalyticsEvent` to use the RTE

Tested locally

#### Checklist:

### DIFF
--- a/packages/io-sign/src/event.ts
+++ b/packages/io-sign/src/event.ts
@@ -1,6 +1,11 @@
 import { IsoDateFromString } from "@pagopa/ts-commons/lib/dates";
 import * as t from "io-ts";
 import * as TE from "fp-ts/lib/TaskEither";
+import * as E from "fp-ts/lib/Either";
+import * as RTE from "fp-ts/lib/ReaderTaskEither";
+import * as L from "@pagopa/logger";
+
+import { pipe } from "fp-ts/lib/function";
 import { Id, newId } from "./id";
 import {
   SignatureRequestDraft,
@@ -12,6 +17,7 @@ import {
   SignatureRequestWaitForQtsp,
 } from "./signature-request";
 import { IssuerEnvironment } from "./issuer";
+import { ConsoleLogger } from "./infra/console-logger";
 
 const EventId = Id;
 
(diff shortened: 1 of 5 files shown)


gunzip (approved):

packages/io-sign/src/event.ts
```diff
@@ -120,11 +126,78 @@ export type SendEvent = (
...
+      sendEvent(eventAnalyticsClient),
+      TE.map(() => signatureRequest),
+      TE.chainFirstIOK(() =>
+        L.debug("Send analytics event", {
+          eventName,
+          signatureRequest,
+        })({
+          logger: ConsoleLogger,
+        })
+      ),
```
lucacavallaro, line 188: here you should use the logger as `RTE` without passing `ConsoleLogger`

packages/io-sign/src/event.ts
```diff
@@ -120,11 +126,78 @@ export type SendEvent = (
...
+      TE.altW(() =>
+        pipe(
+          TE.right(signatureRequest),
+          TE.chainFirstIOK(() =>
+            L.error("Unable to send analytics event", {
+              eventName,
+              signatureRequest,
+            })({
+              logger: ConsoleLogger,
+            })
```
lucacavallaro, line 199: ditto

packages/io-sign/src/event.ts
grausof: here we are inside a TE of the domain of an event. Furthermore, I don't have to show or return the error to the caller but only log it in the console.

lucacavallaro: We can't use "ConsoleLogger" in io-sign, when possible all TE should be converted in RTE

lucacavallaro: ConsoleLogger should be intended only as a "temporary" workaround

grausof: https://github.com/pagopa/io-sign/pull/134

PR #139, opened 2023-04-27
test: [SFEQS-1552] Add unit test create signature request

silvicir:
#### List of Changes
1. Migrated `CreateSignatureRequest` in `io-func-sign-issuer` workspace to `handler-kit@1`
2. Added unit test
3. Updated `openapi.yaml` to return `422` if body is invalid in `/signature-requests POST`

Unit test + manual e2e test in development environment

#### Checklist:


gunzip (commented):
very good!!!

apps/io-func-sign-issuer/src/infra/http/handlers/create-signature-request.ts
```diff
@@ -68,6 +69,9 @@ export const CreateSignatureRequestHandler = H.of((req: H.HttpRequest) =>
       )
     ),
     RTE.chainW(insertSignatureRequest),
+    RTE.chainFirstW((request) =>
+      pipe(EventName.SIGNATURE_CREATED, createAndSendAnalyticsEvent(request))
+    ),
```
gunzip, line 74: I suspect that the order of arguments for `createAndSendAnalyticsEvent` should be swapped. Usually the principle is to put configuration first (`EventName.SIGNATURE_CREATED`) and data (`request`) last. In this case you could have used:

```suggestion
    RTE.chainFirstW(createAndSendAnalyticsEvent(EventName.SIGNATURE_CREATED)),
```

gunzip: see also https://stackoverflow.com/questions/5863128/ordering-of-parameters-to-make-use-of-currying

silvicir: I had already discussed this @lucacavallaro. We decided to take a look at the function `createAndSendAnalyticsEvent` at a later time with @grausof.

grausof: Right! I would say make this change with a separate PR.

grausof (approved):
LGTM

PR #148, opened 2023-05-08
chore: Add missing properties in local.settings.json.example

silvicir:
#### List of Changes

Add missing properties in `local.settings.json.example` in workspace `io-func-sign-user`

### DIFF
--- a/apps/io-func-sign-user/package.json
+++ b/apps/io-func-sign-user/package.json
@@ -49,7 +49,7 @@
     "@pagopa/openapi-codegen-ts": "^12.2.0",
     "@rushstack/eslint-patch": "^1.2.0",
     "@vitest/coverage-c8": "^0.30.1",
-    "azure-functions-core-tools": "^4.0.4895",
+    "azure-functions-core-tools": "^4.0.5148",
     "eslint": "^8.28.0",
     "npm-pack-zip": "^1.3.0",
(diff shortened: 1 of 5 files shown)


apps/io-func-sign-user/package.json
```diff
@@ -49,7 +49,7 @@
     "@pagopa/openapi-codegen-ts": "^12.2.0",
     "@rushstack/eslint-patch": "^1.2.0",
     "@vitest/coverage-c8": "^0.30.1",
-    "azure-functions-core-tools": "^4.0.4895",
+    "azure-functions-core-tools": "^4.0.5148",
```
grausof, line 52: if necessary I would also update it on issuer and support function, otherwise I think that for this PR it may not even be updated.

grausof (changes requested):

gunzip (approved):

apps/io-func-sign-user/package.json
silvicir: I don't know about you but it is necessary on my local machine

grausof: Ok, so we also use the same version on `io-func-sign-issuer` and `io-func-sign-support`

grausof (approved):
LGTM

PR #152, opened 2023-05-11
chore: Code refactor and patch document validation on issuer SDK

grausof:
This PR applies some fixes to the SDK that is being developed by @filippo-tenaglia 
In particular:
- Adds better error handling
- Adds a patch on document upload and validation

Thanks to this SDK a very important BUG of `io-sign` has come out. In particular, it is not possible to upload all the files together (although they have different urls and the API allow it) and then validate them as there is no lock on the DB resources. What you have to do is wait for each single document to be validated before moving on to the next document.

I will do a more detailed description of this bug on confluence.


grausof:
FYI @gunzip @lucacavallaro @silvicir

PR #155, opened 2023-05-16
feat: [SFEQS-1575] Signature request cancelled by issuer

silvicir:
#### List of Changes

1. Added `CANCELLED` as a possible body value for `/signature_requests/{id}/status` endpoint
2. Updated `SetSignatureRequestStatus` function with `handler-kit@1` and added unit tests
3. Updated `io-func-sign-issuer/openapi.yaml` to return `422` if body is invalid in `/signature-requests/{id}/status` `PUT`
4. Added a new function `UpdateSignatureRequest` on `io-func-sign-user` workspace
5. Added a new directory `src/infra/handlers` to contain handlers which are not `http` related on both `io-func-sign-issuer`and `io-func-sign-user` workspaces

#### Motivation and Context

A new signature request status `CANCELLED` is now available. The issuer could now cancel a signature request -with `WAIT_FOR_SIGNATURE` status only.
Flow: The `SetSignatureRequestStatus` function is called with status `CANCELLED`. The function upserts the database and sends a message to the new queue `waiting-for-signature-request-updates`. The queue triggers `UpdateSignatureRequest` function user-side which upserts the database.

Unit test + manual e2e test in development environment

#### Checklist:

Depends on pagopa/io-infra#564
Depends on pagopa/io-infra#565


apps/io-func-sign-issuer/src/infra/http/encoders/signature-request.ts
```diff
@@ -79,6 +79,13 @@ export const SignatureRequestToApiModel: E.Encoder<
           signed_at: extra.signedAt,
         };
       }
+      case "CANCELED": {
+        return {
+          ...commonFields,
+          status: SignatureRequestStatusEnum.CANCELED,
+          canceled_at: extra.canceledAt,
+        };
+      }
```
grausof, line 88: ditto

apps/io-func-sign-user/CancelSignatureRequest/function.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bindings": [
+      {
+        "authLevel": "function",
+        "type": "queueTrigger",
+        "direction": "in",
+        "name": "cancelSignatureRequest",
+        "queueName": "waiting-for-signature-request-cancelation",
```
grausof, line 8:
```suggestion
        "queueName": "waiting-for-signature-request-cancellation",
```
The reference to the infrastructure (io-infra) must be added to this PR to create the new queue

apps/io-func-sign-user/src/infra/http/encoders/signature-request.ts
```diff
@@ -90,6 +89,13 @@ export const SignatureRequestToApiModel: E.Encoder<
           signed_at: extra.signedAt,
         };
       }
+      case "CANCELED": {
+        return {
+          ...commonFields,
+          status: SignatureRequestStatusEnum.REJECTED,
+          reject_at: extra.canceledAt,
+        };
+      }
```
grausof, line 98: The user side must also have the canceled status
```suggestion
      case "CANCELED": {
        return {
          ...commonFields,
          status: SignatureRequestStatusEnum.CANCELLED,
          cancelled_at: extra.canceledAt,
        };
      }
```

packages/io-sign/src/signature-request.ts
```diff
@@ -131,6 +131,19 @@ export type SignatureRequestRejected = t.TypeOf<
   typeof SignatureRequestRejected
 >;
 
+export const SignatureRequestCanceled = makeSignatureRequestVariant(
+  "CANCELED",
+  t.type({
+    canceledAt: IsoDateFromString,
+    qrCodeUrl: t.string,
```
grausof, line 138: If it is canceled I think the qrCode should also be eliminated

apps/io-func-sign-user/src/infra/http/encoders/signature-request.ts
silvicir: sure! it was a mistake from copy and paste, thank you!

packages/io-sign/src/signature-request.ts
silvicir: ok, I was unsure about this

apps/io-func-sign-issuer/src/infra/http/handlers/set-signature-request-status.ts
```diff
@@ -0,0 +1,85 @@
...
+          signatureRequest,
+          markAsReady,
+          RTE.fromEither,
+          RTE.chainW(upsertSignatureRequest),
+          RTE.chainFirstW((req) =>
+            pipe(req, createAndSendAnalyticsEvent(EventName.SIGNATURE_READY))
+          )
+        );
+      } else {
+        return pipe(
```
gunzip, line 73: I'dd add the same check as above or a switch:

`if (body === SetSignatureRequestStatusBodyEnum.CANCELED) {`

apps/io-func-sign-issuer/src/infra/http/handlers/set-signature-request-status.ts
```diff
@@ -0,0 +1,85 @@
...
+          signatureRequest,
+          markAsCanceled(new Date()),
+          RTE.fromEither,
+          RTE.chainW(upsertSignatureRequest)
+        );
+      }
+    }),
+    RTE.chainW(enqueueSignatureRequest),
```
gunzip, line 81: can you add a comment on why we enqueue the signature request here?

packages/io-sign/src/signature-request.ts
lucacavallaro: nope @grausof 

Canceled requests must still be reachable (since they are already reachable in the `message flow`)

apps/io-func-sign-user/CancelSignatureRequest/function.json
lucacavallaro, line 1: imho instead of `CancelSignatureRequest` there should be a generic `UpdateSignatureRequest` function that handles all status update from `issuer` side

gunzip:
next time consider making smaller PRs so they can be approved faster :-)

apps/io-func-sign-issuer/src/app/main.ts
```diff
@@ -100,6 +100,11 @@ const onSignatureRequestReadyQueueClient = new QueueClient(
   "on-signature-request-ready"
 );
 
+const WaitingForSignatureRequestCancelationQueueClient = new QueueClient(
+  config.azure.storage.connectionString,
+  "waiting-for-signature-request-cancellation"
```
lucacavallaro, line 105: maybe we can create a single queue for all status update from issuers

lucacavallaro:
@gunzip in this project we don't use env var for queue names and cosmos containers, only the parent "account/containers" are configurable

apps/io-func-sign-issuer/src/app/main.ts
silvicir: in another PR?

lucacavallaro: nope, you can update this PR

`waiting-for-signature-request-cancellation` -> `waiting-for-signature-request-updates`

`CancelSignatureRequest` -> `UpdateSignatureRequest`

the logic for now should be the same :)

silvicir:
Have you seen the log issue I reported in the description? @gunzip @grausof @lucacavallaro

apps/io-func-sign-user/CancelSignatureRequest/function.json
gunzip: do you need to create the queue using terraform before deploying this code?

grausof: PR here: https://github.com/pagopa/io-infra/blob/main/src/domains/sign/storage.tf

grausof:
> > > Have you seen the log issue I reported in the description? @gunzip @grausof @lucacavallaro
> > 
> > 
> > you can think about adding them. The logs were inserted only in some functions due to lack of time.
> 
> I meant that I can't see logs

Ok sorry, you can set the logging level by adding the following environment variable to the` local.settings.json` file

```
"AzureFunctionsJobHost__logging__logLevel__Function": "Trace"
```

silvicir:
> > > > Have you seen the log issue I reported in the description? @gunzip @grausof @lucacavallaro
> > > 
> > > 
> > > you can think about adding them. The logs were inserted only in some functions due to lack of time.
> > 
> > 
> > I meant that I can't see logs
> 
> Ok sorry, you can set the logging level by adding the following environment variable to the` local.settings.json` file
> 
> ```
> "AzureFunctionsJobHost__logging__logLevel__Function": "Trace"
> ```
> 
> take also a look [here](https://github.com/anthonychu/functions-log-suppression)

Thank you!!!

apps/io-func-sign-issuer/src/infra/http/handlers/set-signature-request-status.ts
```diff
@@ -0,0 +1,98 @@
...
+    validate(SetSignatureRequestStatusBody),
+    E.filterOrElse(
+      (status) => status === "READY" || status === "CANCELLED",
+      () => new Error("only READY or CANCELLED are allowed")
+    )
+  );
+
+const getQueue =
+  (signatureRequest: SignatureRequest) =>
+  (queueClient: { ready: QueueClient; cancelled: QueueClient }) => {
```
grausof, line 42: Why are there still two queues here if they have been merged into one?

apps/io-func-sign-issuer/src/infra/http/handlers/set-signature-request-status.ts
```diff
@@ -0,0 +1,98 @@
...
+    } else if (signatureRequest.status === "CANCELLED") {
+      return E.right(queueClient.cancelled);
+    } else {
+      return E.left(new Error("There are no queues for this update"));
+    }
+  };
+
+const enqueueSignatureRequest =
+  (signatureRequest: SignatureRequest) =>
+  (queueClient: { ready: QueueClient; cancelled: QueueClient }) =>
```
grausof, line 54: ditto

grausof:
‼️ Remember to add the triggered function name `UpdateSignatureRequest` inside [staging_disabled](https://github.com/pagopa/io-infra/blob/47bfeeeeebbedaae4666ca1cc1d41f239b75aaa8/src/domains/sign/io_sign_user_func.tf#L3) configuration to avoid that the function trigger also fires on the staging environment. 

Furthermore, once the terraform has been applied, the check (`deployment slot setting`) must be enabled on the azure portal

@silvicir

apps/io-func-sign-issuer/src/infra/http/handlers/set-signature-request-status.ts
grausof: ok then i would rename the queue here as well!
```suggestion
  (queueClient: { ready: QueueClient; updated: QueueClient }) => {
```

grausof (approved):

PR #159, opened 2023-05-18
fix(issuer): Make signature fields optional in validate-document

lucacavallaro:
#### List of Changes

1. the `validate document` endpoint now it also accept only the pdf document to validate if there are no signature fields to be tracked

#### Motivation and Context

the signature fields are optional, even in the rest API this information can be omitted

unit test

#### Checklist:

### DIFF
--- a/apps/io-func-sign-issuer/openapi.yaml
+++ b/apps/io-func-sign-issuer/openapi.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.3
 info:
   title: Firma con IO - Issuer API
-  version: 1.2.0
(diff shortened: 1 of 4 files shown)


grausof (approved):

apps/io-func-sign-issuer/openapi.yaml
```diff
@@ -326,6 +326,8 @@ paths:
                   type: array
                   items:
                     $ref: "#/components/schemas/SignatureField"
+              required:
```
gunzip, line 329: this is a breaking change, isn't it?

PR #172, opened 2023-06-13
ci: [SFEQS-1603] add GitHub action to validate openapi spec (using the OPEX validator)

manuelcoppotelli:
#### List of Changes

Defining github actions to check opex config file and openapi syntax

#### Motivation and Context

When modifying the swagger file, we want to be sure it will be valid against the specification

#### Checklist:

### DIFF
--- a/.github/workflows/opex-review/action.yaml
+++ b/.github/workflows/opex-review/action.yaml
@@ -1,0 +1,33 @@
+name: opex-review
+description: Generate the dashboard or fail and exit in case of syntax errors
+
+inputs:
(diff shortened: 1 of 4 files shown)


lucacavallaro (changes requested):
can you generalize these workflow files? (using templates, runtime variables..)

manuelcoppotelli:
> can you generalize these workflow files? (using templates, runtime variables..)

I extracted the behavior on a dedicated function template

lucacavallaro (approved):

PR #174, opened 2023-06-26
feat: [SFEQS-1566] Add PDF magic number check to document validation

lucacavallaro:
#### List of Changes

the `validateDocument` function now checks if the uploaded document content starts with `%PDF` (the "magic number" associated with PDF documents) 

#### Motivation and Context

a PDF file, to be valid, should start with `%PDF`. this check doesn't guarantee us that the pdf files are valid, but it excludes malformed uploads (for example the ones created with `multipart/form-data` request that is not supported by our system)

unit tested (`validate document` tests)

#### Checklist:


apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
```diff
@@ -99,18 +100,36 @@ const loggingContext = (meta: UploadMetadata) => ({
 // Checks:
 // 1 - check if the document is compatible with given metadata
 export const validateDocument = (
```
silvicir, line 102: This numbered list in the comments could be a bit confusing

silvicir (changes requested):

apps/io-func-sign-issuer/src/app/use-cases/validate-upload.ts
silvicir: and since the endpoints `validate-document` and `validate-upload` are already written with `handler-kit@1`, maybe we could add a comment at the beginning of each handler (as we said we would do during the refactor)

PR #179, opened 2023-07-06
feat: [SFEQS 1761] Setup "io-sign-backoffice-ca" service

lucacavallaro:
#### List of Changes

1. Add `io-sign-backoffice-ca` workspace in `apps/` folder
2. Setup Next 13 + `app router` + `@pagopa/mui-italia`

⚠️  Since `@pagopa/mui-italia` does not yet support the latest version of `@mui` we are temporarily pinning this dependencies to a specific version

```json
    "@emotion/react": "11.10.5",
    "@emotion/styled": "11.10.5",
    "@mui/icons-material": "5.11.0",
    "@mui/lab": "5.0.0-alpha.119",
    "@mui/material": "5.11.8"
```

#### Motivation and Context

This new workspace will contain the source code of the backoffice app. We chose to use Next as web framework with `app router`.

#### Checklist:


apps/io-sign-backoffice-ca/src/app/_components/Quote.tsx
```diff
@@ -0,0 +1,13 @@
+"use client";
+
+import { use } from "react";
+import { Box, Typography } from "@mui/material";
+
+export default function Quote({ content }: { content: Promise<string> }) {
+  const quote = use(content);
```
lucacavallaro, line 7: TIL: asynchronous client components should not use `await` to block rendering! there is a new `hook` for this specific task: `use`

apps/io-sign-backoffice-ca/next.config.js
```diff
@@ -0,0 +1,11 @@
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  reactStrictMode: true,
+  modularizeImports: {
```
gunzip, line 4: cc @datalek

silvicir (approved):

apps/io-sign-backoffice-ca/src/app/_components/Quote.tsx
gunzip, line 7: why this component should take a promise?

lucacavallaro: because it is an async component, it's how `Suspense` + RSC with streaming works

_...summarizing by omitting some internal details..._

The "container" component (in this case `Home` in `app/page.tsx`) starts the `Promise` (`getRandomQuote`) without waiting for it's completion (no `await` in this step, as you can see at line `11` in `page.tsx`) in this way the rendering is not blocked and it's performed on the server side leaving just an "hole" for a client component (`Quote` in this case) that will be rendered only when the promise (passed as prop) is resolved. In the meantime that the Promise is running, suspense renders a `fallback` component (`<Loading />` in our case)

apps/io-sign-backoffice-ca/src/app/_components/mui/Box.tsx
lucacavallaro, line 1: **TIL**

(1) You can compose `client` and `server` components, BUT `client components` must be explicity marked using the `"use client";` directive. Without this directive, React assumes that the component can be rendered on the server;

(2) RSC does not have support for hooks such as `useRef` or `useState`;

If you compose your RSC with a component that doesn't have the right directive but uses hooks, the build _fails_.

And this is the case of the `Box` component from `@mui/material` it does not yet include the `use client` directive but it contains a call to `useRef` so, without this wrapper is not composable with RSC.

PR #182, opened 2023-07-13
feat(backoffice): [SFEQS-1865] Create api key endpoint - back office

silvicir:
#### List of Changes
1. Added `api/api-keys` route `POST` method to create an api key
2. Added unit tests

#### Motivation and Context
To run the application, you have to execute `az login` first

Trough unit tests with `vitest`.

#### Checklist:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+**/node_modules
+**/dist 
+**/.next
+.git
--- a/apps/io-sign-backoffice-ca/Dockerfile
+++ b/apps/io-sign-backoffice-ca/Dockerfile
@@ -1,0 +1,49 @@
+FROM node:18-alpine AS base
+
+FROM base AS builder
+
+RUN apk add --no-cache libc6-compat
+RUN apk update
+
+WORKDIR /app
+COPY . .
+RUN yarn set version berry
+RUN yarn dlx turbo prune --scope=io-sign-backoffice-ca --docker
+
+# Add lockfile and package.json's of isolated subworkspace
+FROM base AS installer
+RUN apk add --no-cache libc6-compat
+RUN apk update
+WORKDIR /app
+
+# First install the dependencies (as they change less often)
+COPY .gitignore .gitignore
+COPY --from=builder /app/out/json/ .
+COPY --from=builder /app/out/yarn.lock ./yarn.lock
+COPY --from=builder /app/.yarnrc.yml ./yarnrc.yml
+RUN yarn install
+
+# Build the project
+COPY --from=builder /app/out/full/ .
+COPY turbo.json turbo.json
+
+RUN yarn turbo run build --filter=io-sign-backoffice-ca...
+
+FROM base AS runner
+LABEL org.opencontainers.image.source=https://github.com/pagopa/io-sign
+WORKDIR /app
+
+ENV NODE_ENV production
+ENV NEXT_TELEMETRY_DISABLED 1
+
+# Don't run production as root
+RUN addgroup --system --gid 1001 nodejs
+RUN adduser --system --uid 1001 nextjs
+USER nextjs
+
+COPY --from=installer /app/apps/io-sign-backoffice-ca/next.config.js .
+COPY --from=installer --chown=nextjs:nodejs /app/apps/io-sign-backoffice-ca/.next/standalone/node_modules ./node_modules
+COPY --from=installer --chown=nextjs:nodejs /app/apps/io-sign-backoffice-ca/.next/standalone/apps/io-sign-backoffice-ca .
+COPY --from=installer --chown=nextjs:nodejs /app/apps/io-sign-backoffice-ca/.next/static .next/static
+
+CMD node server.js


apps/io-sign-backoffice-ca/.env
```diff
@@ -0,0 +1,10 @@
+SUBSCRIPTION_ID=# override this variable in .env.local
+RESOURCE_GROUP_NAME=# override this variable in .env.local
+SERVICE_NAME=# override this variable in .env.local
+PRODUCT_NAME=# override this variable in .env.local
```
lucacavallaro, line 4: what is the purpose of this variable?

apps/io-sign-backoffice-ca/.env
```diff
@@ -0,0 +1,10 @@
+SUBSCRIPTION_ID=# override this variable in .env.local
+RESOURCE_GROUP_NAME=# override this variable in .env.local
```
lucacavallaro, line 2: can you give a proper prefix to each variable?

ex. `AZURE_` or `COSMOS_`

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/apim.ts
```diff
@@ -0,0 +1,48 @@
...
+      cause: result.error.issues,
+    });
+  }
+  return result.data;
+};
+
+const apimConfig = getApimConfig();
+let apimClient: ApiManagementClient | null = null;
+
+export const getApimClient = () => {
```
lucacavallaro, line 40: you should wrap these kind of functions with `cache` function from `react`

in this way you can avoid the top-level `apimClient` variable

apps/io-sign-backoffice-ca/src/lib/error.ts
```diff
@@ -0,0 +1,36 @@
+export class ParsingInputError extends Error {
```
lucacavallaro, line 1: imho we can use just `ZodError` for parsing/validation error

apps/io-sign-backoffice-ca/src/app/api/api-keys/route.ts
```diff
@@ -0,0 +1,14 @@
...
+const getResponseFromError = (e: Error) => {
+  const { error, status } = getHttpError(e);
+  return NextResponse.json({ error }, { status });
+};
+
+export async function POST(request: NextRequest) {
+  return createApiKey(request)
+    .then((body: unknown) => NextResponse.json(body, { status: 201 }))
+    .catch(getResponseFromError);
+}
```
lucacavallaro, line 14: imho we can manage the response in a more explicit/plain way

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
...
+        throw new Error();
+      }
+      return primaryKey;
+    })
+    .catch((e) => {
+      throw new SubscriptionCreationError({
+        cause: e,
+      });
+    });
+}
```
lucacavallaro, line 133: I think that this function should manage the sync errors (getApimClient and getApimConfig)

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
...
+        await readApiKey(apiKey);
+        return apiKey;
+      })
+      .then(async (apiKey) => {
+        const primaryKey = await createApimSubscription(
+          apiKey.resourceId,
+          apiKey.displayName
+        );
+        return { ...apiKey, primaryKey };
+      })
```
lucacavallaro, line 151: imho we should not mix `then` with `async/await` 

can you try to refactor it using only then or only async/await?

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
+import { z } from "zod";
+import { id as newId } from "@io-sign/io-sign/id";
```
lucacavallaro, line 2: I suggest you to use directly the `ulid` lib instead of `@io-sign/io-sign`

`@io-sign/io-sign` is a big library that makes sense only if it used extensively across the project

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
...
+  CosmosDatabaseError,
+} from "@/lib/cosmos";
+import { ParsingInputError } from "@/lib/error";
+
+const ApiKeyBody = z.object({
+  institutionId: z.string().nonempty(),
+  displayName: z.string().nonempty(),
+  environment: z.enum(["TEST", "DEFAULT", "INTERNAL"]),
+  resourceId: z.string().nonempty(),
+  name: z.string().nonempty(),
```
lucacavallaro, line 20: what is the difference between `name` and `displayName`?

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
...
+import {
+  getCosmosClient,
+  getCosmosConfig,
+  CosmosDatabaseError,
+} from "@/lib/cosmos";
+import { ParsingInputError } from "@/lib/error";
+
+const ApiKeyBody = z.object({
+  institutionId: z.string().nonempty(),
+  displayName: z.string().nonempty(),
```
lucacavallaro, line 17: `displayName` should have a default value, random-generated

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,161 @@
...
+
+type ApiKeyBody = z.infer<typeof ApiKeyBody>;
+
+type ApiKey = ApiKeyBody & {
+  id: string;
+  status: "ACTIVE" | "INACTIVE";
+  createdAt: Date;
+};
+
+const newApiKey = (apiKey: ApiKeyBody): ApiKey => ({
```
lucacavallaro, line 31: (don't kill me!) I think that we should align with standard javascript constructors that are `PascalCased` and without any `prefix`

so this function can be renamed `ApiKey` (like `Number`, `Date` or `Boolean`)

lucacavallaro (changes requested):

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
silvicir: all right. btw it is similar to the other constructors in `io-sign` like `newDossier` or `newSignatureRequest`

apps/io-sign-backoffice-ca/.env
silvicir: as you can see the subscriptions on Azure are associated with a product scope

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
lucacavallaro: @silvicir yes I know! but I think that I made a mistake on `io-sign` :)

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
silvicir: we agreed that `displayName` should be generated front-end side

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -0,0 +1,142 @@
...
+    apiKeyBody.displayName
+  );
+  const apiKey = ApiKey(apiKeyBody);
+  await insertApiKey(apiKey);
+
+  return {
+    id: apiKey.id,
+    key,
+  };
+}
```
lucacavallaro, line 142: 1. Is `readApiKey` really necessary? that is the behaviour of `createApimSubscription` and `insertApiKey` on duplicate input?
2. Imho `readApiKey` (in the case if it's really needed) should be named `getApiKey` and should take only `displayName` and `institutionId` parameters! we should not allow the same displayName in different environments
3. What happen if `createApimSubscription` succeed and `insertApiKey` fails? maybe we should run a "cleanup" job that deletes the api key 
4. I'm thinking we can remove `resourceId` and use just the `id` of `ApiKey` - what do you think?

lucacavallaro (changes requested):

apps/io-sign-backoffice-ca/src/app/api/api-keys/_lib/api-key.ts
silvicir: Right now this is the input:
```
{
    "institutionId": "a0e07d4a-9792-4af3-8175-889aead727b8",
    "displayName": "Comune di C",
    "environment": "DEFAULT",
    "resourceId": "178909225251"
}
```
on duplicated input `createApimSubscription` will not create a new subscription while `insertApiKey` will create a new item.

lucacavallaro (approved):

PR #184, opened 2023-07-21
feat(backoffice): [SFEQS-1885] Implemented authentication from Self Care + minor improvements

lucacavallaro:
#### List of Changes

1. Added `/auth/[id]` and `/auth/logout` routes to manage authentication. The first one verifies the id token passed by self care portal and creates a sealed session cookie (like [the ones in ruby on rails](https://guides.rubyonrails.org/security.html#session-storage) ) based on jwt, the second route destroys the cookie.
2. Improved translation, now they are totally server-side driven. No more `messages.json` in `bundle.js` and delays/FUOC while resolving the translations.
3. Improved the main header with documentation, support and logout button. Added the footer.
4. Added a proper home page!
5. Revamped the project structure, to use a more "idiomatic" style for Next.js and React standards.

With this changes we are now able to deploy safely this service in production :) 

The authentication part is fully unit-tested with jest. There are not E2E tests for now.

#### Screenshots (if appropriate):

image

#### Checklist:

### DIFF
--- a/apps/io-sign-backoffice-ca/package.json
+++ b/apps/io-sign-backoffice-ca/package.json
@@ -20,11 +20,13 @@
     "@types/react-dom": "18.2.6",
     "eslint": "8.44.0",
     "eslint-config-next": "13.4.8",
-    "next": "13.4.8",
-    "next-intl": "^2.17.5",
+    "jose": "^4.14.4",
+    "next": "^13.4.11",
+    "next-intl": "3.0.0-beta.8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "typescript": "5.1.6",
+    "vitest": "^0.33.0",
     "zod": "^3.21.4"
   }
 }
--- a/apps/io-sign-backoffice-ca/src/app/auth/_lib/self-care.ts
+++ b/apps/io-sign-backoffice-ca/src/app/auth/_lib/self-care.ts
@@ -1,0 +1,59 @@
+import { createRemoteJWKSet, jwtVerify } from "jose";
+import { z } from "zod";
+import { cache } from "react";
+
+const Config = z
+  .object({
+    AUTH_SELFCARE_JWK_SET_URL: z.string().url(),
+    AUTH_SELFCARE_JWT_ISSUER: z.string().nonempty(),
+    AUTH_SELFCARE_JWT_AUDIENCE: z.string().nonempty(),
+  })
+  .transform((e) => ({
+    jwks: new URL(e.AUTH_SELFCARE_JWK_SET_URL),
+    audience: e.AUTH_SELFCARE_JWT_AUDIENCE,
+    issuer: e.AUTH_SELFCARE_JWT_ISSUER,
+  }));
+
+const getConfig = cache(() => {
+  const result = Config.safeParse(process.env);
+  if (!result.success) {
+    throw new Error("error parsing auth.self-care config", {
+      cause: result.error.issues,
+    });
+  }
+  return result.data;
+});
+
+const Claims = z.object({
+  // issued at (in ms)
+  iat: z.number(),
+  uid: z.string().uuid(),
+  email: z.string().email(),
+  name: z.string().nonempty(),
+  family_name: z.string().nonempty(),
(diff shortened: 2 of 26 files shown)


apps/io-sign-backoffice-ca/src/app/auth/_lib/self-care.ts
```diff
@@ -0,0 +1,59 @@
+import { createRemoteJWKSet, jwtVerify } from "jose";
+import { z } from "zod";
+import { cache } from "react";
```
lucacavallaro, line 3: TIL the backend code that perform data fetching (from db, from environment, from external backing services...) can use the built-in `cache` function from `react` to memoize their result.

silvicir (approved):

apps/io-sign-backoffice-ca/src/app/auth/_lib/user.ts
```diff
@@ -0,0 +1,64 @@
...
+  } = await verifySelfCareIdToken(idToken);
+  const user = {
+    id,
+    email,
+    firstName,
+    lastName,
+  };
+  const maxAge = (desired_exp - iat) / 1000;
+  await createSessionCookie(user, maxAge);
+  return { user, institutionId };
```
silvicir, line 41: Is `{ institutionId }` not enough?

apps/io-sign-backoffice-ca/src/app/auth/_lib/session.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return cookie;
+}
+
+export async function getPayloadFromSessionCookie(): Promise<{}> {
+  let payload: {};
+  try {
+    const cookie = getSessionCookie();
+    const secret = getSecret();
+    const result = await jwtVerify(cookie.value, secret);
```
silvicir, line 67: Can we try `then` instead of `await` to avoid temporary variables?

apps/io-sign-backoffice-ca/src/app/auth/_lib/session.ts
```diff
@@ -0,0 +1,75 @@
...
+
+const defaultCookieOptions = {
+  secure: true,
+  httpOnly: true,
+  sameSite: true,
+};
+
+const cookieName = "_iosign_session";
+
+const getSecret = cache(() => {
```
gunzip, line 14: what about something like:

```suggestion
const getSecret = cache(({ secret } = process.env) => {
```

in this way you don't have to inspect the method implementation to find out which env you need to mock in tests

lucacavallaro: this is fine as a pattern in case you need to test the functions directly, but this is not the case of `getSecret` that is tested indirectly by testing the other (exported) functions of the module

apps/io-sign-backoffice-ca/src/app/auth/_lib/session.ts
lucacavallaro: the temp variable you want to avoid is `let payload` ?

silvicir: yes

apps/io-sign-backoffice-ca/src/app/auth/_lib/user.ts
lucacavallaro: I'll keep the user for now, because _non si sa mai_ 
I don't want to tie the implementation of this function with the controller

apps/io-sign-backoffice-ca/package.json
```diff
@@ -20,11 +20,13 @@
...
+    "jose": "^4.14.4",
+    "next": "^13.4.11",
+    "next-intl": "3.0.0-beta.8",
     "react": "18.2.0",
     "react-dom": "18.2.0",
     "typescript": "5.1.6",
+    "vitest": "^0.33.0",
     "zod": "^3.21.4"
   }
 }
```
silvicir, line 32: check if any of these dependencies should be devdependencies

PR #188, opened 2023-08-01
added listApiKeys function and unit tests

silvicir:
#### List of Changes

1. Added `listApiKeys` function
2. Added unit tests

Through unit tests

#### Checklist:

### DIFF
--- a/apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
+++ b/apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
@@ -3,10 +3,14 @@ import { ulid } from "ulid";
 import { getApimClient, getApimConfig } from "./apim";
 import { getCosmosClient, getCosmosConfig } from "@/lib/cosmos";
 
+const Environment = z.enum(["TEST", "DEFAULT", "INTERNAL"]);
+
+type Environment = z.infer<typeof Environment>;
+
 export const ApiKeyBody = z.object({
   institutionId: z.string().nonempty(),
   displayName: z.string().nonempty(),
-  environment: z.enum(["TEST", "DEFAULT", "INTERNAL"]),
+  environment: Environment,
 });
 
 type ApiKeyBody = z.infer<typeof ApiKeyBody>;
@@ -31,30 +35,13 @@ export class ApiKeyAlreadyExistsError extends Error {
   }
 }
 
-async function getApiKey(displayName: string, institutionId: string) {
+async function apiKeyExists(institutionId: string, displayName: string) {
   try {
-    const { cosmosDbName, cosmosContainerName } = getCosmosConfig();
-    const cosmosClient = getCosmosClient();
-    const { resources } = await cosmosClient
-      .database(cosmosDbName)
-      .container(cosmosContainerName)
-      .items.query({
-        parameters: [
-          {
-            name: "@displayName",
-            value: displayName,
-          },
-          {
-            name: "@institutionId",
-            value: institutionId,
-          },
-        ],
-        query:
-          "SELECT * FROM c WHERE c.displayName = @displayName AND c.institutionId = @institutionId",
-      })
-      .fetchAll();
-    if (resources.length !== 0) {
-      throw new ApiKeyAlreadyExistsError();
+    const apiKeys = await getApiKeys(institutionId, { displayName });
+    if (apiKeys.length !== 0) {
+      throw new ApiKeyAlreadyExistsError(
+        "such name already exists for the institution"
+      );
     }
   } catch (e) {
     throw e instanceof ApiKeyAlreadyExistsError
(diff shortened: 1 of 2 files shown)


apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -124,9 +83,138 @@ async function deleteApimSubscription(subscriptionId: string) {
...
+      ? e
+      : new Error("unable to create the API key", { cause: e });
+  }
+}
+
+async function insertApiKey(apiKey: ApiKey) {
+  try {
+    await getCosmosContainer().items.create(apiKey);
+  } catch (e) {
+    await deleteApimSubscription(apiKey.id);
```
gunzip, line 105: Is this some kind of rollback? Maybe it would be better to handle this in the caller than as a side effet of insertApiKey
```suggestion
    await deleteApimSubscription(apiKey.id);
```

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -124,9 +83,138 @@ async function deleteApimSubscription(subscriptionId: string) {
...
+      throw new ApiKeyAlreadyExistsError(
+        "such name already exists for the institution"
+      );
+    }
+  } catch (e) {
+    throw e instanceof ApiKeyAlreadyExistsError
+      ? e
+      : new Error("unable to create the API key", { cause: e });
+  }
+}
```
lucacavallaro, line 99: I would simplify this!

#### Option 1



```typescript
import assert from "node:assert"

async function apiKeyExists(institutionId: string, displayName: string) {
  try {
    const apiKeys = await getApiKeys(institutionId, { displayName });
    assert.equal(apiKeys.length === 0, true);
  } catch {
    throw new ApiKeyAlreadyExistsError()
   }
}
```

#### Option 2

```typescript
async function apiKeyExists(institutionId: string, displayName: string) {
  try {
    const apiKeys = await getApiKeys(institutionId, { displayName });
    return apiKeys.length === 0;
  } catch {
    return false;
   }
}
```

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -2,12 +2,16 @@ import { z } from "zod";
...
 import { getApimClient, getApimConfig } from "@/lib/apim";
-import { getCosmosClient, getCosmosConfig } from "@/lib/cosmos";
+import { getCosmosContainer } from "@/lib/cosmos";
+
+const Environment = z.enum(["TEST", "DEFAULT", "INTERNAL"]);
+
+type Environment = z.infer<typeof Environment>;
 
 export const ApiKeyBody = z.object({
   institutionId: z.string().nonempty(),
```
lucacavallaro, line 12:
```suggestion
  institutionId: z.string().uuid(),
```

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -32,66 +36,20 @@ export class ApiKeyAlreadyExistsError extends Error {
...
-      .items.create(apiKey);
-  } catch (e) {
-    await deleteApimSubscription(apiKey.id);
-    throw new Error("unable to create the API key", {
-      cause: e,
-    });
-  }
-}
-
 async function createApimSubscription(resourceId: string, displayName: string) {
```
lucacavallaro, line 39:
```suggestion
async function createApimSubscription(id: string, displayName: string) {
```

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
silvicir: in this way if `getApiKeys` threw an error I would throw `ApiKeyAlreadyExistsError `

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -124,9 +83,138 @@ async function deleteApimSubscription(subscriptionId: string) {
...
+
+          return {
+            id,
+            key: primaryKey ?? "",
+            institutionId,
+            displayName,
+            environment,
+            status,
+            createdAt,
+          };
```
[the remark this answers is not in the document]
lucacavallaro: ok, so we should parse it with `ApiKey.parse` to remove the unwanted properties!

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
lucacavallaro: I don't think it's a problem

apps/io-sign-backoffice-app/src/app/api/api-keys/_lib/api-key.ts
```diff
@@ -32,108 +47,144 @@ export class ApiKeyAlreadyExistsError extends Error {
...
-async function insertApiKey(apiKey: ApiKey) {
+async function apiKeyExists(
+  institutionId: string,
+  displayName: string
+): Promise<boolean> {
+  const apiKeys = getApiKeys(institutionId, { displayName });
+  let exists = false;
+  for await (const _apiKey of apiKeys) {
+    exists = true;
+  }
```
lucacavallaro, line 118: here you can simply call `apiKeys.next()` and check its content

if it returns 
```js
{ value: undefined, done: true }
```

there are no items, so you should return `false`

otherwise it returns

```js
{ value: (... your ApiKey object ...), done: false }
```

so you know that there are elements, and you can return `true`

--- 

your refactored function might look like this

```ts
const apiKeys = getApiKeys(institutionId, { displayName });
const item = apiKeys.next();
return !item.done;
```

lucacavallaro (approved):

PR #190, opened 2023-08-28
chore(backoffice): Removed environment filter in api keys list

silvicir:
#### Checklist:


lucacavallaro (approved):

gunzip:
lacks "Motivation and Context"

silvicir:
the keys will be displayed in two different tabs, one for each environment. this way the front end will fetch all the keys and we will avoid the loading when switching the tab (since the keys will be few)

PR #191, opened 2023-08-29
chore(backoffice): Replace arm SDK with API Management REST API

silvicir:
#### List of Changes

The `@azure/arm-apimanagement` and `@azure/identity` dependencies have been removed. Now the API management REST API with access token is used.

Through unit tests

#### Checklist:

### DIFF
--- a/apps/io-sign-backoffice-app/src/lib/apim.ts
+++ b/apps/io-sign-backoffice-app/src/lib/apim.ts
@@ -1,6 +1,4 @@
 import { z } from "zod";
-import { DefaultAzureCredential } from "@azure/identity";
-import { ApiManagementClient } from "@azure/arm-apimanagement";
 import { cache } from "react";
 
 const Config = z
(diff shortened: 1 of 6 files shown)


lucacavallaro (approved):

apps/io-sign-backoffice-app/src/lib/apim.ts
```diff
@@ -31,58 +31,95 @@ class ApimProductClient {
...
 
-  async getSecret(id: string): Promise<string | undefined> {
-    const { primaryKey } = await this.client.subscription.listSecrets(
-      this.resourceGroupName,
-      this.serviceName,
-      id
-    );
-    return primaryKey;
+  async getSecret(id: string): Promise<string> {
+    const url = `https://iosignbackofficedev.management.azure-api.net/subscriptions/${this.subscriptionId}/resourceGroups/${this.resourceGroupName}/providers/Microsoft.ApiManagement/service/${this.serviceName}/subscriptions/${id}/listSecrets?api-version=2022-08-01`;
```
gunzip, line 44: What about encapsulating this template string into its own method?

gunzip (commented):
Just note that the drawback of this approach is that the access token is unique for all REST API access.

silvicir:
> Just note that the drawback of this approach is that the access token is unique for all REST API access.

@lucacavallaro

PR #196, opened 2023-09-06
feat(bo): [SFEQS-1983] Added get api key endpoint

silvicir:
#### List of Changes

1. Added endpoint GET `/api-keys/[api-key]` to retrieve the api key by id. If the query param `include` with value `institution` is present, the institution's info will be given as well
2. Added `getIssuer` function
3. Refactored `ApimProductClient`class

Through unit tests

#### Checklist:


apps/io-sign-backoffice-app/src/lib/apim.ts
```diff
@@ -21,105 +21,88 @@ const Config = z
...
+        Authorization: this.#accessToken,
       },
-      method: "POST",
+      method,
+      body,
     });
-    if (!res.ok) {
-      throw new Error("error getting primary key on apim");
+    if (!resp.ok) {
+      throw new Error("Something went wrong");
```
gunzip, line 65: I'd try to pass a more explicit error from the caller otherwise debugging would be hard.

apps/io-sign-backoffice-app/src/lib/api-keys/cosmos.ts
```diff
@@ -84,3 +83,27 @@ export async function upsertApiKeyField<
...
+          {
+            name: "@id",
+            value: id,
+          },
+        ],
+        query: "SELECT * FROM c where c.id = @id",
+      })
+      .fetchAll();
+    const apiKey = apiKeySchema
+      .array()
```
lucacavallaro, line 101: `array` of what?

apps/io-sign-backoffice-app/src/lib/apim.ts
silvicir: yes I agree, but this is a temporary code since probably we will switch back to arm

apps/io-sign-backoffice-app/src/lib/api-keys/cosmos.ts
lucacavallaro: nvm

lucacavallaro (approved):

PR #198, opened 2023-09-07
feat(bo): [SFEQS-1989] Add "Create Issuer" endpoint

lucacavallaro:
#### List of Changes

1.  changed `Issuer` shape to better fit with `Cosmos DB` schema model
2. added a new `/api/issuers` API endpoint for creating new `Issuers`
3. refactored `openapi.json` to include only the information used by Azure API Management
4. add 404 response to `/api/api-keys` endpoint

#### Motivation and Context

This is a porting of the `Issuer` entity from `io-func-sign-issuers` service to `io-sign-backoffice-app` service. `Issuer` is a specialization of the `Institution` resource from self care, which contains all information specific to the `io-sign` domain.

unit tested

#### Checklist:


apps/io-sign-backoffice-app/src/lib/issuers/use-cases.ts
```diff
@@ -1 +1,30 @@
-export { getIssuer } from "./cosmos";
+import { ulid } from "ulid";
+import { z } from "zod";
+
+import { Institution } from "@/lib/institutions";
+
+import { issuerSchema } from "./index";
+import { insertIssuer, getIssuer } from "./cosmos";
+
+export const createIssuerPayloadSchema = issuerSchema.pick({
```
silvicir, line 9: Maybe we should move `createIssuerPayloadSchema` and `CreateIssuerPayload` in `index.ts`

silvicir (approved):

PR #221, opened 2023-10-05
[SFEQS-2023] Add io-sign-backoffice-func

silvicir:
#### List of Changes

1. Created `io-sign-backoffice-func` directory
2. Added `onSelfcareContractsMessage` function with Event Hubs trigger

#### Checklist:


apps/io-sign-backoffice-func/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "io-sign-backoffice-func",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "build": "tsup-node",
+    "start": "func start"
```
lucacavallaro, line 7: can you add here `prettier` and `eslint` ? this package json should contain `lint`, `lint:fix` and `format` script

lucacavallaro (changes requested):

gunzip:
can we add a readme with the setup to make this run locally and the description of the function?

silvicir:
> can we add a readme with the setup to make this run locally and the description of the function?

Then we should add a readme for all the functions. The command is just `yarn run start` so maybe we could update the `io-sign` README.

gunzip:
It's ok to put this information where we can find all the descriptions, one for each function.

lucacavallaro:
> can we add a readme with the setup to make this run locally and the description of the function?

I think this approach can lead us to bad practices and poor documentaiton.

The instructions to run the project locally are (and must be!) the same for all the workspaces. 

In particular, each workspace must have its own `build`, `start`, `format` and `lint` script. No workspace should have exceptions and different ways of working.

Maybe we should update the `README` to enforce this pattern and remove all the references to specific workspaces.

PR #223, opened 2023-10-09
chore: [SFEQS-2024] Add state property to issuer

silvicir:
#### List of Changes

Add issuer `state` property when issuer is created in `io-sign-backoffice-app` and `io-func-sign-issuer`

#### Motivation and Context

An issuer could withdraw from the contract and we have to handle this case.  So we need a issuer property `state` with values `active` and `inactive`.

#### Checklist:


lucacavallaro (approved):
it's ok but we have to update all `issuer` documents on db to avoid causing disruptions

silvicir:
> it's ok but we have to update all `issuer` documents on db to avoid causing disruptions

it's already on db

PR #230, opened 2023-10-20
chore(bo): [SFEQS-2008] Send slack notification

lucacavallaro:
#### List of Changes

1. Added `sendMessage` that sends slack messages
2. Used `sendMessage` to notify first access to portal and api key creation

#### Motivation and Context

Our product team needs to be notified promptly about certain backoffice events

#### Checklist:

### DIFF
--- a/apps/io-sign-backoffice-app/src/lib/issuers/use-cases.ts
+++ b/apps/io-sign-backoffice-app/src/lib/issuers/use-cases.ts
@@ -2,8 +2,9 @@ import { ulid } from "ulid";
 
 import { Institution } from "@/lib/institutions";
 
-import { CreateIssuerPayload } from "./index";
+import { CreateIssuerPayload, issuerSchema } from "./index";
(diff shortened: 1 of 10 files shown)


silvicir (approved):

apps/io-sign-backoffice-app/src/lib/issuers/use-cases.ts
```diff
@@ -2,21 +2,26 @@ import { ulid } from "ulid";
...
     }
-    await insertIssuer({
+    const newIssuer = issuerSchema.parse({
       externalId: ulid(),
       type: "PA",
       state: "active",
       ...payload,
     });
+    await insertIssuer(newIssuer);
+    await sendMessage(
```
gunzip, line 22: if `sendMessage` fails the issuer is created but we return an error. what about switching to "fire and forget" mode for this kind of logging operations instead?

lucacavallaro: `sendMessage

PR #237, opened 2023-10-30
chore(bo): [SFEQS-2025] Migrate issuer onboarding slack message

silvicir:
#### List of Changes

When a new institution onboards, the function `CreateIssuer` on `io-func-sign-issuer` gets triggered and a slack message is sent. We wanted to move this logic in the back office domain

#### Checklist:


apps/io-sign-backoffice-func/src/app/main.ts
```diff
@@ -0,0 +1,51 @@
...
+const configOrError = pipe(
+  getConfigFromEnvironment(process.env),
+  E.getOrElseW(identity)
+);
+
+if (configOrError instanceof Error) {
+  throw configOrError;
+}
+
+const config = configOrError;
```
lucacavallaro, line 22: I would remove fp-ts code here, since this portion of code throws.
imho it can be just a zod parse

apps/io-sign-backoffice-func/src/app/main.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const issuerRepository = new BackOfficeIssuerRepository(
+  config.backOffice.apiBasePath,
+  config.backOffice.apiKey
+);
+
+const institutionRepository = new SelfcareInstitutionRepository(
+  config.selfcare.api.basePath,
+  config.selfcare.api.key
+);
```
lucacavallaro, line 32: these repositories can be composed in one `BackofficeApiClient` class

apps/io-sign-backoffice-func/src/app/main.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const slackRepository = new SlackMessageRepository(config.slack.webhookUrl);
```
lucacavallaro, line 34: I would replace this class with just a `sendSlackMessage` function (maybe we can reuse the backoffice one?)

to me it's an overkill defining a class just for one operation

apps/io-sign-backoffice-func/src/infra/back-office/config.ts
lucacavallaro, line 1: it can be replaced entirely by zod, I don't see advantages of `fp-ts` here (but I know why you implemented it in this way)

apps/io-sign-backoffice-func/src/infra/back-office/issuer.ts
```diff
@@ -0,0 +1,77 @@
...
+import { parse } from "../handlers/handler-kit/validation";
+
+const issuer = z.object({
+  id: z.string().min(1),
+  type: z.enum(["PA"]),
+  externalId: z.string().min(1),
+  institutionId: z.string().uuid(),
+  supportEmail: z.string().email(),
+  status: z.enum(["active", "inactive"]),
+});
```
lucacavallaro, line 19: if this schema must be identical to the backoffice one, move it to `@io-sign/io-sign`

apps/io-sign-backoffice-func/src/infra/handlers/__test__/on-selfcare-contracts-message.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+  Institution,
+  InstitutionRepository,
+} from "@/infra/selfcare/institution";
+import { Issuer, IssuerRepository } from "@/infra/back-office/issuer";
+import { SlackRepository } from "@/infra/slack/message";
+
+describe("onSelfcareContractsMessage handler", () => {
+  let issuerRepository: IssuerRepository;
+  let institutionRepository: InstitutionRepository;
+  let slackRepository: SlackRepository;
```
lucacavallaro, line 17: use `vitest` [context](https://vitest.dev/guide/test-context.html) instead of `describe-scoped` variable

apps/io-sign-backoffice-func/src/infra/handlers/handler-kit/handler-kit-azure-func.ts
lucacavallaro, line 1: this file should be moved to `pagopa/io-std`

apps/io-sign-backoffice-func/src/infra/handlers/health.ts
```diff
@@ -0,0 +1,5 @@
+import { HttpResponseInit } from "@azure/functions";
+
+export async function healthHandler(): Promise<HttpResponseInit> {
+  return { status: 200 };
```
lucacavallaro, line 4: empty body = `204`

apps/io-sign-backoffice-func/src/infra/selfcare/contract.ts
lucacavallaro, line 1: to me this file can be simplified, we really need `baseContract` ? I think we can keep only a `contractScheme` scheme with the `io-sign` contract

apps/io-sign-backoffice-func/src/infra/selfcare/use-cases.ts
```diff
@@ -0,0 +1,19 @@
+import { pipe } from "fp-ts/lib/function";
+import * as TE from "fp-ts/lib/TaskEither";
+import * as O from "fp-ts/lib/Option";
+import { InstitutionEnvironment } from "./institution";
+
+export const getSupportEmail =
```
lucacavallaro, line 6: `supportEmail` must be read only from `issuer`

the `supportEmail` in institution it's not reliable, because it can't be updated

lucacavallaro (changes requested):

apps/io-sign-backoffice-func/src/infra/selfcare/use-cases.ts
silvicir: we only have this `supportEmail` at this stage
btw it is just an onboarding message, it is not supposed to be reliable

apps/io-sign-backoffice-app/src/app/api/api-keys/[api-key]/route.ts
```diff
@@ -6,7 +6,7 @@ import { NextRequest, NextResponse } from "next/server";
 import { z } from "zod";
 
 const pathSchema = z.object({
-  "api-key": z.string().nonempty(),
+  "api-key": z.string().min(1),
```
Comment (gunzip) [inline apps/io-sign-backoffice-app/src/app/api/api-keys/[api-key]/route.ts:9]: why this change?

apps/io-sign-backoffice-func/src/infra/selfcare/use-cases.ts
```diff
@@ -0,0 +1,19 @@
...
+  (internalInstitutionId: string) =>
+  ({ institutionRepository }: InstitutionEnvironment) =>
+    pipe(
+      internalInstitutionId,
+      institutionRepository.getById.bind(institutionRepository),
```
gunzip, line 11:
```suggestion
      institutionRepository.getById(internalInstitutionId),
```

apps/io-sign-backoffice-func/src/infra/slack/message.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  constructor(webhookUrl: string) {
+    this.#webhookUrl = webhookUrl;
+  }
+
+  sendMessage(text: string): TE.TaskEither<Error, void> {
+    return pipe(
+      TE.tryCatch(
+        () =>
+          fetch(this.#webhookUrl, {
```
gunzip, line 25: all those fetches should have a timeout (and use keepalive ideally, see https://learn.microsoft.com/en-us/azure/app-service/troubleshoot-intermittent-outbound-connection-errors#cause)

apps/io-sign-backoffice-func/src/infra/selfcare/use-cases.ts
lucacavallaro: we don't need to have `supportEmail` in the message (we don't use it), so can be removed at all

apps/io-sign-backoffice-func/src/infra/selfcare/contract.ts
```diff
@@ -0,0 +1,43 @@
+import { z } from "zod";
+
+const contract = z.object({
```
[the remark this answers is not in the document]
lucacavallaro: I saw that you made a discriminated union on `state` (`ioSignContracts` schema) but is this necessary? maybe `contractSchema` alone is enough

apps/io-sign-backoffice-func/src/infra/slack/issuer-message.ts
```diff
@@ -0,0 +1,13 @@
...
+  description,
+}: {
+  internalInstitutionId: string;
+  vatNumber: string;
+  description: string;
+}) =>
+  `👋🏻 A new issuer has just been onboarded to io-sign 🚀!\n
+    Description: ${description}\n
+    institutionId: \`${internalInstitutionId}\`\n
+    vatNumber: \`${vatNumber}\`\n`;
```
lucacavallaro, line 13: here you can use the same "format" as newer slack messages sent by the backoffice

something like

> (_backoffice_) __DESCRIPTION__ (`INSTITUTION_ID`) ha effettuato l'onboarding.

packages/io-sign/src/issuer.ts
```diff
@@ -27,7 +28,17 @@ export const Issuer = t.type({
...
+
+// this is the back office issuer schema
+export const issuerSchema = z.object({
+  id: z.string().min(1),
+  type: z.enum(["PA"]),
+  externalId: z.string().min(1),
+  institutionId: z.string().uuid(),
+  supportEmail: z.string().email(),
+  status: z.enum(["active", "inactive"]),
+});
```
lucacavallaro, line 44: maybe this schema can be shared

apps/io-sign-backoffice-func/src/infra/http/fetch.ts
```diff
@@ -0,0 +1,15 @@
+import { Agent, RequestInfo } from "undici";
+
+export const fetchWithTimeoutAndKeepAlive = (
```
lucacavallaro, line 3: instead a new `fetch` wrapper, I would export only the configured dispatcher

so when I need this config I can use in this way `fetch(url, { dispatcher, .... }`

apps/io-sign-backoffice-func/src/infra/slack/use-cases.ts
```diff
@@ -0,0 +1,6 @@
+import { SendMessage } from "./message";
+
+export const sendMessageToSlack =
+  (message: string) =>
+  ({ sendMessage }: SendMessage) =>
+    sendMessage(message);
```
lucacavallaro, line 6: I think we can use directly the `sendMessage` function from the `message` module

lucacavallaro (changes requested):

packages/io-sign/src/issuer.ts
silvicir: what do you mean?

apps/io-sign-backoffice-func/src/app/main.ts
```diff
@@ -0,0 +1,27 @@
...
+import { azureFunction } from "@pagopa/handler-kit-azure-func";
+import { getConfigFromEnvironment } from "./config";
+import { healthHandler } from "@/infra/handlers/health";
+import { onSelfcareContractsMessageHandler } from "@/infra/handlers/on-selfcare-contracts-message";
+import { ioSignContracts } from "@/infra/selfcare/contract";
+import { IoTsType } from "@/infra/handlers/validation";
+import { getById } from "@/infra/back-office/issuer";
+import { sendMessage } from "@/infra/slack/message";
+
+const { backOffice, slack, selfcare } = getConfigFromEnvironment();
```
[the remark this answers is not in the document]
silvicir: an error is thrown

apps/io-sign-backoffice-func/src/infra/selfcare/contract.ts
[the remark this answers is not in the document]
silvicir: I used a discriminated union so I could use the type guard `isActive`

lucacavallaro (approved):

PR #245, opened 2023-12-29
Fix minor syntax errors

nardil:
exclusiveMaximum specifies the exclusive maximum value accepted, requiring an integer value. If exclusiveMaximum is present, maximum becomes redundant.

minLength is not applicable in array definition; minItems is the appropriate one.

The version has been incremented from 1.2.1 to 1.2.2.

#### Checklist:


nardil:
Added missing `maxLength` to `DossierTitle`: according to [user manual](https://docs.pagopa.it/manuale-operativo-di-firma-con-io/creare-il-dossier) the title length must be less then 75 chars

apps/io-func-sign-issuer/openapi.yaml
```diff
@@ -436,8 +436,7 @@ components:
             The HTTP status code generated by the origin server for this
             occurrence of the problem.
           minimum: 100
-          maximum: 600
-          exclusiveMaximum: true
+          exclusiveMaximum: 600
```
lucacavallaro, line 439: @nardil according to OpenAPI 3.0.x documentation, `exclusiveMaximum` requires a boolean value (ref: https://swagger.io/docs/specification/data-models/data-types/#range). I noticed that according to the latest JSON Schema specification the behaviour of this field has changed, but OpenAPI uses a different version of OpenAPI (see https://json-schema.org/understanding-json-schema/reference/numeric#range).

So for now we can't accept this change. If it's a problem for your integration, feel free to reach us at firmaconio-tech@pagopa.it and we will figure out a better solution for your needs.

apps/io-func-sign-issuer/openapi.yaml
```diff
@@ -543,6 +542,7 @@ components:
     DossierTitle:
       type: string
       minLength: 3
+      maxLength: 75
```
lucacavallaro, line 545: Thanks for this edit! Good catch!

To approve it, however, we need to refactor the code, in order to introduce runtime validation and a specific error on the API. We have scheduled this development for next week, you will be credited as a contributor.

lucacavallaro:
Hi @nardil thank you for your contribution 🎉 I left some comments to your changes, feel free to reply directly to those threads.

apps/io-func-sign-issuer/openapi.yaml
nardil: You're right: I got distracted by the new Swagger validator, which actually flags it as an error (https://editor-next.swagger.io/).

PR #254, opened 2024-02-08
[SFEQS-2066, SFEQS-2107] Update Yarn, pipeline fix and improvements

lucacavallaro:
#### List of Changes
1. Update `yarn@3` to `yarn@4` (same API, less dependencies that can interfere with turbo)
2. Add a tempfix to a `turbo prune` bug that broke our `backoffice-deploy` pipeline (add `@babel/types` as dev dependencies).
3. Remove `tag` trigger from deploy workflows since due to GitHub limitations (a workflow can't trigger another workflows) has never worked
4. Now `issuer-func` deploy add a `latest` tag to make the latest deployed version easy to be referenceable in our public documentation

#### Motivation and Context

__(2)__ 
`turbo prune` (used in our pipelines) generates a new `yarn.lock` file with only the dependencies required by our selected workspace. It works fine with all workspaces but `backoffice-app` when, due to a strange incompability between `yarn` and `emotion` (used by `mui`),  it generates a broken `yarn.lock` that breaks our deploy pipeline. Now I applied a tempfix but this bug will be explored further in the future.

__(1)__
I updated `yarn` as an attempt to resolve (2). It didn't work, but the updated `yarn` has the same API of the previous version so I decided to keep this version.

#### Checklist:


.github/workflows/code-review.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
             ${{ runner.os }}-turbo-
 
       - name: "Generate API models"
-        run: yarn workspaces foreach run generate:api-models
+        run: yarn workspaces foreach --all run generate:api-models
```
silvicir, line 48: why the change?

silvicir (approved):

.github/workflows/deploy-issuer-func.yaml
```diff
@@ -14,3 +11,19 @@ jobs:
       environment: io-p-sign-issuer-func
       workspace-name: io-func-sign-issuer
     secrets: inherit
+
+  # tag the current commit as "latest" in order to make the latest
+  # released version easily referenceable in the documentation
+  add_tag_latest:
+    needs: deploy_workspace_to_azure
+    steps:
```
silvicir, line 19: Is `runs-on` property missing?

.github/workflows/code-review.yaml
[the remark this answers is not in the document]
lucacavallaro: `yarn workspaces` in `yarn 4.x` is no longer an external plugin, but it is included directly. By including it they introduced small syntax changes. `yarn workspaces foreach --all` is the same of the old `yarn workspaces foreach`, `--all` is now required.

PR #256, opened 2024-02-09
[SFEQS-1861] Track REJECTED requests on appinsights

lucacavallaro:
#### List of Changes
1. Deprecated `MarkAsSigned` and `MarkAsRejected` functions in favor of `CloseSignatureRequest`
2. Add `closeSignatureRequest` use case that unifies all the "end" statuses of the signature request
3. Rewrote io-services, pdv, analytics code to make it compatible with RTE
4. Integrated Applicationinsights to track REJECTED requests

#### Motivation and Context
tracking the REJECTED event is crucial for us to proactively open assistance ticker to our main upstream dependency (QTSP)

#### Checklist:


apps/io-func-sign-issuer/MarkAsRejected/function.json
```diff
@@ -10,5 +10,5 @@
     }
   ],
   "scriptFile": "../dist/main.js",
-  "entryPoint": "MarkAsRejected"
+  "entryPoint": "CloseSignatureRequest"
```
gunzip, line 13: why this change with the same directory name?

apps/io-func-sign-issuer/MarkAsSigned/function.json
```diff
@@ -10,5 +10,5 @@
     }
   ],
   "scriptFile": "../dist/main.js",
-  "entryPoint": "MarkAsSigned"
+  "entryPoint": "CloseSignatureRequest"
```
gunzip, line 13: ditto

apps/io-func-sign-issuer/src/signature-request-notification.ts
```diff
@@ -61,3 +71,20 @@ export const makeSendSignatureRequestNotification =
...
+  (
+    buildNotificationMessage: (request: SignatureRequest) => NotificationMessage
+  ) =>
+  (request: SignatureRequest) =>
+    pipe(
+      getFiscalCodeBySignerId(request.signerId),
+      RTE.chainW((fiscalCode) =>
+        submitNotification(fiscalCode, buildNotificationMessage(request))
+      ),
+      RTE.map(() => true),
```
gunzip, line 86: what about `bimap`?

packages/io-sign/src/infra/azure/appinsights/config.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const getApplicationInsightsConfigFromEnvironment: RE.ReaderEither<
+  NodeJS.ProcessEnv,
+  Error,
+  ApplicationInsightsConfig
+> = sequenceS(RE.Apply)({
+  instrumentationKey: readFromEnvironment("APPINSIGHTS_INSTRUMENTATIONKEY"),
+  samplingPercentage: pipe(
+    readFromEnvironment("APPINSIGHTS_SAMPLING_PERCENTAGE"),
+    RE.chainEitherKW(t.number.decode),
```
gunzip, line 24: this is supposed decodes a string into a number?

gunzip (commented):
what's the reason to use classes instead of `Readers`?

packages/io-sign/src/infra/azure/appinsights/config.ts
lucacavallaro: yes, maybe I can restrict the target runtype (integer, from 0 to 100)

lucacavallaro:
> what's the reason to use classes instead of `Readers`?

There are not used instead of Readers, the code was refactored only to use Readers in a proper way :) 

I use classes to write the concrete implementation of the dependencies

apps/io-func-sign-issuer/src/signature-request-notification.ts
lucacavallaro: `altW` does not map left value, it allow us to "recover from an error" transforming a `Left` in a `Right`

the result of this function is not `Either<false, true>` but `Either<never, false | true>` - to make it cleaner I think I can use `fold` instead to transform this `RTE` in a `RT<boolean>` since for our perspective it does not fails.

lucacavallaro: done! ✅

apps/io-func-sign-issuer/src/app/main.ts
```diff
@@ -138,6 +180,16 @@ export const MarkAsSigned = makeRequestAsSignedFunction(
   eventAnalyticsClient
 );
 
+export const CloseSignatureRequest = CloseSignatureRequestFunction({
```
silvicir, line 183: should we delete `MarkAsSigned` and `MarkAsRejected` functions?

silvicir (approved):

packages/io-sign/src/infra/io-services/message.ts
```diff
@@ -118,3 +120,20 @@ export const makeSubmitMessageForUser =
...
+  #ioApiClient: IOApiClient;
+  #configurationId: Ulid;
+
+  constructor(ioApiClient: IOApiClient, configurationId: Ulid) {
+    this.#ioApiClient = ioApiClient;
+    this.#configurationId = configurationId;
+  }
+
+  submit(fiscalCode: FiscalCode, notification: NotificationMessage) {
+    return makeSubmitMessageForUser(
```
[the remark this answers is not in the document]
lucacavallaro: `makeSubmitMessageForUser` is still used in `sendNotification` handler (which is built using the old version of `handler-kit`) so it can't be removed now.

Since the implementation of `submit` method is identical, I choose to reuse `makeSubmitMessageForUser` to keep the code DRY.

When `sendNotification` will be ported to `handler-kit@1` we can remove this old function and maintain only the `submit` method.

apps/io-func-sign-issuer/src/signature-request.ts
```diff
@@ -54,6 +55,15 @@ export const SignatureRequest = t.union([
 
 export type SignatureRequest = t.TypeOf<typeof SignatureRequest>;
 
+// A request is defined "CLOSED" when its status is one of: CANCELLED, SIGNED, REJECTED.
+// Closed requests cannot be modified, whatever its outcome.
+export const ClosedSignatureRequest = t.union([
+  SignatureRequestSigned,
+  SignatureRequestRejected,
+]);
```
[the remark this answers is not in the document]
lucacavallaro: I removed the reference to `CANCELLED` status. When `closeSignatureRequest` use case will be enhanced to support that status, we can simply update this codec.

PR #287, opened 2024-05-15
fix: set taxCode in sc-contracts as optional field

silvicir:
#### List of Changes

The `taxCode` property in the `contractSchema` and `institutionDetailSchema ` have become optional

#### Motivation and Context

In the message arriving on the `sc-contracts` queue, the `taxCode` property in `institution` has become optional

#### Checklist:


gunzip:
wouldn't this return `null` for existing records?

silvicir:
> wouldn't this return `null` for existing records?

what do you mean?

gunzip:
> > wouldn't this return `null` for existing records?
> 
> what do you mean?

existing records have `taxCode` as id, so when you'll call `issuerAlreadyExists({ id: internalIstitutionID })` won't it return null?

PR #303, opened 2024-09-10
Add license

libremente:
#### List of Changes

This PR:
* adds a `LICENSE` file (MPL-2.0);
* adds licensing info in the `package.json` file (LMK if this makes sense for you)

#### Checklist:

- [x] My change requires a change to the documentation.


libremente:
@lucacavallaro I added `LICENSE` files and the `license` key in the various `package.json`.
The declaration inside `package/io-sign` is a bit redundant considering that we are already declaring the same license info in the root's `package.json`, however I would keep it like this since I believe it's better to explicit the LICENSE conditions for each app/package.
If it's OK for you please squash and merge to avoid unnecessary history noise. 
Thanks!

libremente:
@lucacavallaro any news on this PR? Thanks

PR #309, opened 2024-10-07
[CES-112] Fix Private Endpoint definition for Cosmos DB

Krusty93:
#### List of Changes

Fix the private endpoint definition by adding the missing sub-resource property

#### Motivation and Context

Private endpoints can't work without sub-resource definition. This PEP is defined in ITN region for the io-sign Cosmos DB, but will not be active until the failover won't happen

#### Checklist:


christian-calabrese (approved):
LGTM

PR #313, opened 2024-11-06
[CES-476] Added new storage account iopsignst for ITN migration

cfcuffari120400:
### Motivation and Context

Storage accounts must be replicated to italy north in view of the infrastructure migration

### Major Changes

Create replication of iopsignst in italy north


Civolo:
Dead PR

PR #332, opened 2025-05-15
[IOPLT-1066] Upgrade packages

robbcocco:
#### List of Changes
Upgraded lint config and packages

#### Motivation and Context
This change was needed to upgrade the stack

Tested by running yarn code-review

#### Checklist:


lucacavallaro (dismissed):
Hi @robbcocco,

To make these changes easier to review, I’d like to ask you to separate them into two different PRs: one for the `Node.js` update, and another one for `prettier`, `eslint`, etc.

From a quick glance, I can already point out a few things that should be changed:
1. The latest version of node is `20.19.2`
2. You must not change the `prettier` version or its configuration unless there are compatibility issues, which you must explain and document.
3. We have a shared, battle-tested `eslint` configuration: https://www.npmjs.com/package/@pagopa/eslint-config. It is compatible with `eslint@8`, designed to be "automatic" and mostly "auto-fixable", so there's no need to define custom rules.
4. `type` must not be replaced with `interface` — they behave differently, and using `type` here was intentional.

robbcocco:
> Hi @robbcocco,
> 
> To make these changes easier to review, I’d like to ask you to separate them into two different PRs: one for the `Node.js` update, and another one for `prettier`, `eslint`, etc.

> **I'll make a separate PR for the changes regarding only the Node.js version.**

> 
> From a quick glance, I can already point out a few things that should be changed:
> 
> 1. The latest version of node is `20.19.2`

> **I've upgraded the requirements and the version**

> 2. You must not change the `prettier` version or its configuration unless there are compatibility issues, which you must explain and document.

> **Upgrading to the latest `@pagopa/eslint-config` requires prettier 3.x. The change is not necessary in itself (upgrading nodejs only should not break compatibility) but I thought it was a nice to have among others libraries upgrades**

> 3. We have a shared, battle-tested `eslint` configuration: https://www.npmjs.com/package/@pagopa/eslint-config. It is compatible with `eslint@8`, designed to be "automatic" and mostly "auto-fixable", so there's no need to define custom rules

> **The latest version (4.x) introduces new rules (and likely replaces or removes others), so with @BurnedMarshall we came up with a set of custom rules to minimize the changes migrating from version 3.x to 4.x. To ease the transition in this monorepo, we also decided to make an internal package from which the projects will inherit the configuration**

> 4. `type` must not be replaced with `interface` — they behave differently, and using `type` here was intentional.

> **Continuing on point 3, I've added a rule to block this "fix" from the linter and reverted the changes**

Hi @lucacavallaro, thanks for the feedback.
I've yet to fix some of the issues you reported, in the meantime I've replied inline with my considerations on some of the biggest changes and other issues i've tackled.

apps/io-func-sign-issuer/package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@azure/cosmos": "^4.0.0",
     "@azure/event-hubs": "^5.9.0",
+    "@azure/functions": "^3.5.0",
```
BurnedMarshal, line 28: Double check why this dependency is moved from `devDependencies` in here. If only types are used it should remain as devDependency. There are other cases because you promote it as dependency?

apps/io-func-sign-issuer/src/infra/back-office/issuer.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
```
BurnedMarshal, line 2:
```suggestion
```
Disable this rule locally, not globally for the file.

BurnedMarshal (changes requested):
Only a couple of comments.

apps/io-func-sign-issuer/package.json
robbcocco: Moved back to devDependencies in ff08dcdc4b3432549c28ff9ca32f68d2ed2da4e5

BurnedMarshal (approved):
LGTM

PR #585, opened 2026-04-29
Reject POST /signatures requests with SPID level lower than L3

Civolo:
#### List of Changes

- Added `requireSpidLevel` decoder in `signer.ts` that reads the `x-iosign-spid-level` header and returns a `403 Forbidden` if the header is absent or contains a SPID level lower than L3
- Integrated the decoder into `CreateSignatureHandler` via `sequenceS`, so the check is enforced before any other business logic on `POST /signatures`

#### Motivation and Context

SPID authentication comes in three levels (L1, L2, L3). Qualified electronic signatures require strong authentication, so `POST /signatures` must only accept requests authenticated at SPID level L3. This change enforces that constraint at the HTTP layer.

Manual testing via the existing local dev setup.

#### Screenshots (if appropriate):

N/A

#### Checklist:

Resolves IEL-420

Depends on #584

Depends on https://github.com/pagopa/io-backend/pull/1321


apps/io-func-sign-user/src/infra/http/decoders/signer.ts
```diff
@@ -1,11 +1,43 @@
...
+      () => new H.HttpForbiddenError("Missing x-iosign-spid-level header")
+    ),
+    E.chainW(
+      H.parse(
+        SpidLevel,
+        "The content of x-iosign-spid-level is not a valid SPID level"
+      )
+    ),
+    E.chainW((spidLevel) =>
+      spidLevel === "https://www.spid.gov.it/SpidL3"
```
giuseppedipinto: This check is very strict: it's absolutely fine, but what do you think if we relaxed the check to a less restrictive `contains("spidl3")` after lowercased the value?
This way, we wouldn't be forced to check the previous URL _(e.g., https://www.spid.gov.it/)_, which we have no control over.
It's not a real change request, but rather, a challenge.

Civolo: That's a fair point, and I appreciate the challenge! However, I'd keep the strict enum as is. These exact URIs are part of the SPID technical specification defined by AGID ([official docs.](https://docs.italia.it/italia/spid/spid-regole-tecniche/it/34.1.2/single-sign-on.html)) As such, this enum is already well-established and used consistently across the IO platform. Since the values are institution-defined and not free-form strings, a contains check would be looser than necessary and could mask malformed headers. The URL itself is effectively a stable identifier, not an endpoint we call.

giuseppedipinto: @Civolo I totally agree with you, keep it up! 🙌

Iustin-Burlacu (approved):

giuseppedipinto (approved):

apps/io-func-sign-user/src/infra/http/decoders/signer.ts
```diff
@@ -1,11 +1,43 @@
...
+        ? E.right(spidLevel)
+        : E.left(
+            new H.HttpForbiddenError(
+              "A minimum SPID level of L3 is required to create a signature"
+            )
+          )
```
BurnedMarshal:
```suggestion
    E.chainW(
      E.fromPredicate(
        (spidLevel) => spidLevel === "https://www.spid.gov.it/SpidL3",
        () =>
          new H.HttpForbiddenError(
            "A minimum SPID level of L3 is required to create a signature"
          )
      )
```

apps/io-func-sign-user/src/infra/http/decoders/__test__/signer.spec.ts
```diff
@@ -30,3 +30,81 @@ describe("requireSignerId", () => {
...
+    const req = H.request("my-url");
+    expect(requireSpidLevel(req)).toEqual(
+      expect.objectContaining({
+        left: expect.objectContaining({
+          name: "HttpError",
+          status: 403,
+        }),
+      })
+    );
+  });
```
BurnedMarshal: You can use a more compact way to declare this tests with the same output. It's only a stylish suggestion to reduce code

```suggestion
  it.each([
    {
      label: "SPID level is L1",
      headers: { "x-iosign-spid-level": "https://www.spid.gov.it/SpidL1" },
    },
    {
      label: "SPID level is L2",
      headers: { "x-iosign-spid-level": "https://www.spid.gov.it/SpidL2" },
    },
    { label: "header is missing", headers: {} as Record<string, string> },
  ])("fails with a 403 error when $label", ({ headers }) => {
    const req: H.HttpRequest = {
      ...H.request("my-url"),
      headers,
    };
    expect(requireSpidLevel(req)).toEqual(
      expect.objectContaining({
        left: expect.objectContaining({
          name: "HttpError",
          status: 403,
        }),
      })
    );
  });
```

BurnedMarshal (approved):
LGTM, only some code style suggestion

apps/io-func-sign-user/src/infra/http/decoders/__test__/signer.spec.ts
Civolo: Thx, applied in 738cf28

apps/io-func-sign-user/src/infra/http/decoders/signer.ts
Civolo: Thx, applied in 738cf28

PR #598, opened 2026-05-06
Enable public network access on ITN storage account

Civolo:
## Problem

SAS URL access to the ITN storage account (`iopitnsignst01`) was failing for external clients (mobile apps) because `publicNetworkAccess` was set to `Disabled`. Azure drops the connection before evaluating the SAS token, so authentication never gets a chance to succeed.

## Solution

Replace the `pagopa-dx/azure-storage-account/azurerm` DX module with native `azurerm_*` resources to decouple public network access control from private endpoint management.

The DX module had an internal coupling: setting `force_public_network_access_enabled = true` simultaneously set `allow_nested_items_to_be_public = true` and **destroyed both private endpoints** (blob and queue). This would have broken private DNS routing for internal Function Apps.

### Resulting configuration

- `public_network_access_enabled = true` — SAS URLs reachable from external clients ✅
- `allow_nested_items_to_be_public = false` — no anonymous access ✅
- Both blob and queue private endpoints preserved — Function Apps continue routing via private DNS ✅

### Terraform state migration

25 `moved` blocks cover all resources (storage account, 4 containers, 16 queues, 2 private endpoints, 1 metric alert) for a zero-downtime migration from the DX module state paths to native resource paths.

The terraform plan shows only `~` (in-place updates) for DX tag removal (`ModuleName`, `ModuleSource`, `ModuleVersion`) — no `destroy` on any data-bearing resource.

> ⚠️ The `moved` blocks must be removed in a follow-up commit after the first successful apply.

### DIFF
--- a/infra/italynorth/_modules/storage/data.tf
+++ b/infra/italynorth/_modules/storage/data.tf
@@ -16,3 +16,13 @@ data "azurerm_monitor_action_group" "common_error_action_group" {
   name                = "${local.prefix}${local.env_short}error"
   resource_group_name = "${local.prefix}-${local.env_short}-rg-common"
 }
+
+data "azurerm_private_dns_zone" "storage_blob" {
+  name                = "privatelink.blob.core.windows.net"
+  resource_group_name = "${local.prefix}-${local.env_short}-rg-common"
+}
+
+data "azurerm_private_dns_zone" "storage_queue" {
+  name                = "privatelink.queue.core.windows.net"
+  resource_group_name = "${local.prefix}-${local.env_short}-rg-common"
+}
--- a/infra/italynorth/_modules/storage/main.tf
+++ b/infra/italynorth/_modules/storage/main.tf
@@ -1,63 +1,151 @@
-module "sign_storage_account" {
-  source  = "pagopa-dx/azure-storage-account/azurerm"
-  version = "~> 2.0"
-
-  environment = {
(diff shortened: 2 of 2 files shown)


StefanoDoc (approved):

gunzip:
question: the DX module for the storage account includes a `delegated_access` use case to enable access via SAS token. could this be a possible solution?

https://registry.terraform.io/modules/pagopa-dx/azure-storage-account/azurerm/latest#use-cases-comparison

Civolo:
> question: the DX module for the storage account includes a `delegated_access` use case to enable access via SAS token. could this be a possible solution?
> 
> https://registry.terraform.io/modules/pagopa-dx/azure-storage-account/azurerm/latest#use-cases-comparison

I looked into the \`delegated_access\` use case in the module source before opening this PR.

The name is misleading, but it is designed for **OAuth/Entra ID delegated access**, not SAS-based access.

gunzip:
As far as I know, it’s the application (e.g., an Azure Function) that needs to authenticate using an EntrID identity (with a specific role) to obtain the SAS token, which is then passed to the user. The user can then access the file without Entra ID authentication. Let us double-check that this is correct and we’ll update you.

gunzip:
I confirm that the 'delegated_access' use case allows the application to obtain a SAS token to share with the user, enabling them to access the content online.

